PR #131, opened 2013-10-10
plugins/bind9_rndc: add incoming queries graph

presbrey:
Adds graph for incoming queries by parsing existing rndc output (via multigraph).


leeclemens:
I believe this PR is obsolete since https://github.com/munin-monitoring/munin/commit/562726d53f2f321820006a7614bbf3fd81c0e7d2 was accepted was accepted and merged into master and this PR should be closed

ssm:
It looks like the rewrite of that plugin in #190 added about the same features.

Thanks for the pull request.

PR #158, opened 2014-01-21
Autodetect HZ value in linux/cpu plugin

grooverdan:
Autodetection is better than guessing. The /proc/config more reliable than /boot/config* (Fedora)


ssm:
The "cpu" plugin was, with a handful of others, removed in favour of the procfs multigraph plugin.

Thanks for the pull request.

PR #204, opened 2014-04-03
master: adding a static version of Night&Day

steveschnepp:
Note that the RRD CDEF patterns are taken from a Cacti post

### DIFF
--- a/master/lib/Munin/Master/GraphOld.pm
+++ b/master/lib/Munin/Master/GraphOld.pm
@@ -941,6 +941,7 @@ sub process_service {
     my %total_neg;
     my $autostacking = 0;
 
+    my $first_def;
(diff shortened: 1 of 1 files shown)


steveschnepp:
This needs a option to enable/disable it. Bonus would be to be able to specify the ranges, as it is currently very CET centric.

I don't think it should be per-plugin, as it might blurs the meaning

ssm:
Add a default in Munin::Common::Defaults, and a configuration option to enable it.

kenyon:
What does this do?

steveschnepp:
It does put a not-white background on night periods, in order to identify daily off-patterns

sumpfralle:
Is this change still interesting for us?

kenyon:
> Is this change still interesting for us?

I think it would be a cool feature.

PR #235, opened 2014-07-08
Adding the storable2sql tool

steveschnepp:
Tool to convert 2.0.x Storables to 2.2.x SQL

Its main purpose is to be able to drop the munin-json in place on a vanilla 2.0 install.


steveschnepp:
I dump our IRC chat with @ssm here for the record :

    <ssm> should https://github.com/munin-monitoring/munin/pull/235 be rebased, or closed?  (Adding the storable2sql tool)
    <TheSnide> closed. as i won't have time to finish it.
    <TheSnide> it was a great idea, but it would require too much effort to finish it. and well... i prefer to focus on 2.2

PR #292, opened 2014-11-02
munin_stats plugin: only graph munin-graph if graph_strategy=cron (Close...

steveschnepp:
...s: #767032)


steveschnepp:
I have merged #328, which seems to be more complete, but not fully complete. Anyway, we have to move forward, so closing this.

PR #305, opened 2014-11-12
Use configured plugin group as primary group

knorrie:
Instead of always having the munin group as real and effective group
while executing a plugin, and only using configured group(s) as
additional groups, use the configured groups instead.

When specifying group in plugin-conf, I expect munin to set this group
as the primary group. Maybe I even want to set a group for my plugin to
prevent it from being able to access anything that is owned by the group
that the munin-node daemon is running at!

I stumbled into this issue when writing a plugin that reads from
/proc/<pid>/smaps. When setting the user and group in the plugin config,
access was still denied, because the real/effective user got set to the
specified user, but the group did not get changed. I guess this is a bit
of an edge case, because in most of the cases just having the specified
group available as additional group will "work".

This change will no longer automagically include the munin group itself
in the group list when executing a plugin if it's not expicitely added
in that list. IMHO, this should be the expected behaviour.

### DIFF
--- a/lib/Munin/Node/Service.pm
+++ b/lib/Munin/Node/Service.pm
@@ -185,10 +185,11 @@ sub _resolve_gids
 
     # Support running with more than one group in effect. See documentation on
     # $EFFECTIVE_GROUP_ID in the perlvar(1) manual page.  Need to specify the
-    # default group twice: once for setegid(2), and once for setgroups(2).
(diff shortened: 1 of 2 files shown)


ssm:
The tests fail with this commit. It should be fixed before this can be merged.

If the code is OK, please update the tests

If the commit broke something, please update the code.

knorrie:
The tests reflect the new behaviour now.

PR #319, opened 2014-12-05
Master/Node.pm: added compression to SSH command to drastically reduce n...

wintermancer:
...etwork traffic for munin_async connections. Useful for many hosts with lot of data on slow connections.
This issue came up for us while transferring bigger chunks of data with munin_async over large distances.

### DIFF
--- a/master/lib/Munin/Master/Node.pm
+++ b/master/lib/Munin/Master/Node.pm
@@ -95,7 +95,7 @@ sub _do_connect {
 		return 0;
 	}
     } elsif ($uri->scheme eq "ssh") {
-	    my $ssh_command = "ssh -o ChallengeResponseAuthentication=no -o StrictHostKeyChecking=no ";
+	    my $ssh_command = "ssh -o ChallengeResponseAuthentication=no -o StrictHostKeyChecking=no -C ";
 	    my $user_part = ($uri->user) ? ($uri->user . "@") : "";
 	    my $remote_cmd = ($uri->path ne '/') ? $uri->path : "";
 


ssm:
First step would be to document how it can be done now in the munin guide.

Second step is to look at adding more attributes in the node definitions (compress = true, for instance)

gersilex:
I don't see a reason why this could not be the default behavior, as the underlying ssh option is fully transparent to the application layer.

We can implement this and open a new issue for implementing and (only then) documenting an option to disable it on demand (which I don't see a reason for, as the stream is already altered by the encrpytion of ssh and cannot be captured anyways) if the user wishes this.

Even for very weak processors the compression will work quite well.

ssm:
Configuration changes can be remarkably simple.

Example "~/.ssh/config", If the munin master runs OpenSSH:

    Host *
    Compression yes

Code changes need some thought.

* Measure the performance impact and gains of compression on a 1, 10, 100 and 1000.  Look at numbers, and see if it is worth doing.
* Research command line options for how to do this on the various SSH clients used across operating systems.
* Introduce munin master configuration options at node and global scope
* Adding command line options where applicable.

ssm:
The 2.1.12 milestone is _obviously_ for the documentation changes. :)

steveschnepp:
I'm highly in favor of `.ssh/config`, as it's very easy to configure.

And then I don't really like to add a gazillion options, so it should to be somewhat generic, such as : 

    ssh_options = -o Compress=yes -o CompressionLevel=9

PR #325, opened 2015-01-03
multips: reject 'autoconf' unless $names is set

bjornfor:
Without this, the plugin will accept a non-working (empty) config at
configure time and will exit with error each time it is run for real.

The 'multips_memory' plugin already handles it this way.


bjornfor:
Ping. Please have a look at this small bugfix (and apply it).

bjornfor:
Thanks!

PR #328, opened 2015-01-14
Update munin_stats plugin to work with cgi strategies

leeclemens:
Several updates, including dynamic determination if graph stat should be displayed, adds dependency on File::ReadBackwards, also matches munin-html config generation time when html_strategy cgi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ branches:
   - master
 install: "/bin/true"
 before_install:
+ - "cpanm --notest File::ReadBackwards"
  - "cpanm --notest File::Copy::Recursive"
  - "cpanm --notest File::Slurp"
  - "cpanm --notest HTML::Template"
--- a/plugins/node.d/munin_stats.in
+++ b/plugins/node.d/munin_stats.in
@@ -34,18 +34,18 @@ my @logs = qw/update graph html limits/;
 my $logdir = ($ENV{'logdir'} || $ENV{'MUNIN_LOGDIR'} || '@@LOGDIR@@');
 
 
-sub uses_graph_cgi {
+sub uses_graph_cron {
     my $log_file = $_[0];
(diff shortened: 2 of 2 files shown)


steveschnepp:
First, thanks for the PR :+1: 

It's just that this has been proposed on IRC already, so it seems it's a common idea. I'll write my comments here also for future reference then.

Yet, I don't think that plainly removing the munin_graph field is a good idea. As I still want to be ablet to track the time spend in the CGI part of it, and if one enables munin_graph only on odd days, he'll loose the amount spend on it on even days. (... I know! But, it's just an example).

Adding 2 fields to the graph showing the time spend on the CGI counterparts, along with renaming the munin_graph "cron CGI", and showing it effectively spends 0 time in it seems a more compete move.

leeclemens:
Interesting point about leaving munin_graph as "cron CGI", but what do you mean by being able to track the time spent in the CGI part of it (paragraph above)? Perhaps logging level related, but my munin-cgi-graph.log does not include generation times.

leeclemens:
Also worth noting, using the ReadBackwards module, I fixed a bug I was seeing with the state file and not being able to run munin-run munin_stats ad-hoc (as it overwrote the position).

leeclemens:
Worth noting, I don't plainly remove the munin_graph field - it is conditionally based upon its existence. Your PR does, however, seem to remove it. My suggested fix to this issue would remove it if it is not used - assuming someone isn't switching between cgi and cron every 5 minutes. As mentioned above, I support printing the cron CGI time, but I am not sure where that information is available.

steveschnepp:
I agree on "Your PR does, however, seem to remove it". That's even why that PR isn't merged yet :grinning:

leeclemens:
@steveschnepp what can I do to improve my proposed solution to this issue?

steveschnepp:
Bah. I'm just merging it. We'll refine it later.

PR #329, opened 2015-01-19
Master report pages design enhancements

chteuchteu:
### Fact

Munin is a great tool, but its design hasn't evolved in a while. For my first contribution, I'd like to open for discussion a small design refresh!

Here it is: if anyone has any suggestion or comment about this redesign, don't hesitate to tell!

### Changes
During this redesign, I kept the main page structure unchanged. No HTML has been edited and every part of the interface is where it was before, for two reasons:

1. Changing user habits is usually not a good idea
2. I'd like to avoid breaking the whole HTML structure for my first pull request

The main change you will instantly notice is that the header bar color changed. It is a hard decision, but I think that munin would benefit from having a stronger visual appearance. If this is too much for now (I could totally understand), we could try to reduce the color saturation in order to make it less flashy.

Also, you will notice that the logo had to be edited a bit (hence the new logo-h-neg.png file) to fit the new header color.

Since a picture is worth a thousand words, here's one:
screenshot from 2015-01-19 22 06 21

Good news:

- The report page is now wider (95% vs 90%)
- It is compatible with older web browsers. IE8 just won't display special effects (gradients and border-radius)
- This won't take longer to load. Every effect is 100% CSS and 0% images
- It is prettier, IMO

Bad news:

- There aren't

### And now?

This is not the final version: as said before, this is open for discussion and everyone if welcome to tell what he thinks about it! Also, there may be some more enhancements to be done (CSS file cleanup for example).
I'm also planning to work a bit on the dynazoom page (HTML cleanup and & bit of CSS) in the future.


steveschnepp:
:+1:

ssm:
:+1: 

A very nice improvement on the classic munin interface.

PR #333, opened 2015-01-21
Fix fofields always having the same entries as ofields.

driskell:
It seems that ofields and fofields in the notification system text always list the same fields.

An example is the "df" graph, which has a field per disk. Assuming 4 disks, if one of the disks was in a WARNING state and moved to an OK state, the graph itself is marked as "state_changed". This triggers the email notification - but there is no way to know which field was fixed, ofields contains all OK fields (which in this case is all fields) - and so does fofields. In the code I can see it always sets them identical, first via `push @{$stats{$field->{'state'}}}, $fname;` and then via the check for `'ok'` which populates fofields.

This patch fixes this and makes fofields contain only entries that had a state change to OK. This seems to be the intended functionality (I assume fo = fixed ok?)

I've not been able to test this completely yet though. I copied and pasted the changes from my changes to 2.0.25 which I have installed from RedHat packages. But it should work OK.

### DIFF
--- a/master/lib/Munin/Master/LimitsOld.pm
+++ b/master/lib/Munin/Master/LimitsOld.pm
@@ -222,13 +222,13 @@ Options:
     --force		Alias for \"--always-send ok,warning,critical,unknown\".
                         Overrides --always-send command line, as well as the
                         always_send contact configuration options.
-    --service <service>	Limit notified services to <service>. Multiple 
+    --service <service>	Limit notified services to <service>. Multiple
     			--service options may be supplied.
-    --host <host>	Limit notified hosts to <host>. Multiple --host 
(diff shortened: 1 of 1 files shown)


steveschnepp:
Also, you can even target the stable-2.0 branch for bugfixes, that way it will be in the next 2.0.x.

driskell:
Absolutely. Thanks.

It's Atom - it has a really bad habit of "automatically repairing" whitespace. I'll undo the changes in Git though. I'll also rebase and modify the PR for stable-2.0 - failing that I'll reopen.

driskell:
I've also successfully tested on 2.0 so yes, it'll much wiser that, sorry!

steveschnepp:
Just reopen. I don't think you can change the target. And PR creation is cheap :cake:

PR #334, opened 2015-01-22
Fix fofields always having the same entries as ofields

driskell:
(Resubmitted version of #333 based on stable-2.0)

It seems that ofields and fofields in the notification system text always list the same fields.

An example is the "df" graph, which has a field per disk. Assuming 4 disks, if one of the disks was in a WARNING state and moved to an OK state, the graph itself is marked as "state_changed". This triggers the email notification - but there is no way to know which field was fixed, ofields contains all OK fields (which in this case is all fields) - and so does fofields. In the code I can see it always sets them identical, first via push @{$stats{$field->{'state'}}}, $fname; and then via the check for 'ok' which populates fofields.

This patch fixes this and makes fofields contain only entries that had a state change to OK. This seems to be the intended functionality (I assume fo = fixed ok?)

I now run this in production successfully.

### DIFF
--- a/master/lib/Munin/Master/LimitsOld.pm
+++ b/master/lib/Munin/Master/LimitsOld.pm
@@ -315,6 +315,7 @@ sub process_service {
     $hash->{'group'} = get_full_group_path($hparentobj);
     $hash->{'worst'} = "ok";
     $hash->{'worstid'} = 0 unless defined $hash->{'worstid'};
+    $hash->{'recovered'} = {};
 
     my $state_file = sprintf ('%s/state-%s-%s.storable', $config->{dbdir}, $hash->{group}, $host);
     DEBUG "[DEBUG] state_file: $state_file";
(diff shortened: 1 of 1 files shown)


driskell:
Sorry about the slight misalignment - there's a huge mesh of spaces amongst the indentation in this file so my commits look to be slightly offset due to Github's tab width being different to mine. I can do another commit to remove all space indentation after if it helps.

steveschnepp:
That would be awesome. Just rebase and force-push to your branch. It will fix the PR.

driskell:
Is that OK? I had to bite my lip and use spaces >.<
Looks like the code is a complete mess as it's been pushed to a 4 width indentation using spaces, and the tabs have to be 8 width (like GitHub) to look correct. Should be completely replaced with tabs but that's a separate problem I guess!

steveschnepp:
Thx

driskell:
Thanks - I'll open another shortly for the develop branch so the next version will benefit too. If you're planning to cherry in there manually, let me know. There will be one conflict though on the recovery check but it's easy to resolve.

steveschnepp:
No need, since i merge stable-2.0 regularly in devel. (well, just after each release)

PR #337, opened 2015-01-23
Convert the CGI in a HTTP server

steveschnepp:
CGI are quite cumbersome to configure. HTTP proxiing is now mainstream.


ssm:
This makes me hopeful :)

PR #342, opened 2015-02-07
Add fail2ban_ wildcard plugin for extended status

leeclemens:
This new plugin utilizes new status information added for the next release of fail2ban. It can also create the same IPs per jail graph as the current fail2ban plugin (provided the newer fail2ban version requirement).


ssm:
Merged to plugins/security in the "contrib" repo

https://github.com/munin-monitoring/contrib/commit/8d8f8343aed80dcfafe90d8e1ef85be66d3084f9

Thank you.

leeclemens:
Makes sense, thanks @ssm

PR #346, opened 2015-02-09
Feature/remove preprocessing

ssm:
Pull request for tracking build refactoring.  Not necessarily ready for merge yet.


ssm:
Looks like some plugins fail with perl <= 5.12, so far.

PR #348, opened 2015-02-11
Fix ofields previous state detection not working correctly

driskell:
Hello

Minor fix to a bad logic check.

Jason


driskell:
Hi @ssm - if you need anything from me, let me know.

It'll be good to get this merged and merged through to devel too.
I can see the previous PR never got merged to devel - do you need a separate PR for that? Or is the LimitsOld.pm in there redundant?

Thanks

PR #356, opened 2015-03-05
Calculate diskspace/inode usage percent accurately

AMDmi3:
Instead of using rounded one from df

Also, sort df output as it's output is not sorted by default which makes graph harder to read.


ssm:
Rebased and merged.  Thank you.

PR #359, opened 2015-03-06
remove --logarithmic form disk latency graphs

roock:
The disk latency graphs has the parameter --logarithmic which makes it unreadable (see attached images).

### DIFF
--- a/plugins/node.d.linux/diskstats.in
+++ b/plugins/node.d.linux/diskstats.in
@@ -844,7 +844,7 @@ sub do_config_device {
             print <<"EOF";
 multigraph ${plugin_name}_latency.$graph_id
 graph_title Average latency for /dev/$pretty_device
-graph_args --base 1000 --logarithmic
+graph_args --base 1000
 graph_vlabel seconds
 graph_category disk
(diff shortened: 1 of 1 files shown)


kenyon:
Can't see screen shot.

ssm:
The use of --logarithmic makes it readable.  See, for instance:

image

ssm:
…forgot to say that when you have a graph where the different data points are magnitudes of order away from eachother, logarithmic display is a way to make the graph readable.  Otherwise you'll get the biggest data point visible, and everything else squashed at the bottom.

roock:
I'm sorry the upload of the screenshot did not work correctly. Thank you for the explanation of what the logarithmic option means. In theory it makes sense to use the option, but in practice, for me a sysadmin the graphs are pretty useless with the use of logarithmic. I'm attaching the examples:

When using logarithmic the graphs look like this: 
screen shot 2015-03-06 at 12 19 42
And without logarithmic the graphs look correct:
screen shot 2015-03-08 at 19 22 09
Also take a look at the graph configuration for the overview of the disk latency graph in line 715 (https://github.com/munin-monitoring/munin/blob/devel/plugins/node.d.linux/diskstats.in#L715). It doesn't have the logarithmic set. It should be set there too or removed in the detail graph, but it should be consistent.

PR #361, opened 2015-03-10
backport http_loadtime from devel

roidelapluie:
The http_loadtime is not working in 2.0. It has a syntax error. This
commit backports the changes from development in stable-2.0 so the next
2.0 release will have a working http_loadtime plugin.


steveschnepp:
When backporting, having the original SHA in the commit msg would be awesome.

PR #362, opened 2015-03-13
Busted limits again

driskell:
I had some issues I've fixed in Limits.

- Inheritance of warning/critical now works correctly and does not break subsequent limits
If you had say, mongo_lag graph, and you set warning/critical as follows to apply them to all fields in the graph, it would not work and limits would be ignored:
mongo_lag.warning :43200
mongo_lag.critical :86400
Additionally, in the DEBUG output I could see it corrupted all limits checks afterwards. It would actually add the "host" node as a service to check, which would populate graph_title into the "host" node. This would then corrupt every limit check for that host afterwards because it would mean other checks, when called get_host_node, would end up grabbing the root node.

- Aliased graph fields now obey limits assigned to them
The jmx_threads check has aliased fields. In the limits check it does not process this correctly and does not pass the correct path to munin_get_rrd_filename - thus it cannot grab the current value to check limits.

One thing that might need factoring in here is the field_order duplicates. I can see GraphOld.pm handles these duplicates itself but maybe it could do with abstracting into the munin_get_field_order...

By the way - I'm not sure how much of this is legacy and due to be replaced... I can see lots of work in devel branch and of course the filename is, after all, LimitsOld.pm. Though there's no new Limits.pm in the devel branch. I'm just wondering what the plan is - rewrite ground up? Or just fixup?

### DIFF
--- a/master/lib/Munin/Master/LimitsOld.pm
+++ b/master/lib/Munin/Master/LimitsOld.pm
@@ -161,7 +161,19 @@ sub process_limits {
     my $work_array = [];
     foreach my $workfield (
         @{munin_find_field_for_limits($config, qr/^(critical|warning)/)}) {
-        my $parent = munin_get_parent($workfield);
+        my $parent;
+        if (defined $workfield->{'graph_title'}) {
+            # Limit is defined on a service and inherits to all fields, queue it
+            $parent = $workfield;
+        } else {
+            # Limit is defined on a field, or a non-existent service
+            # Assume field, grab parent service, and verify it is valid
+            $parent = munin_get_parent($workfield);
+            if (!defined $parent->{'graph_title'}) {
+                DEBUG "[DEBUG] Ignoring work item for non-existent service: " . munin_get_node_name($parent) if ($DEBUG);
+                next;
+            }
+        }
         if (!defined $work_hash_tmp{$parent}) {
 	    $work_hash_tmp{$parent} = 1;
 	    push @$work_array, $parent;
@@ -297,7 +309,7 @@ sub process_service {
     my $hparentobj = munin_get_parent($hobj);
     my $parent     = munin_get_node_name($hobj);
     my $gparent    = munin_get_node_name($hparentobj);
-    my $children   = munin_get_children($hash);
+    my $field_order = munin_get_field_order($hash);
 
     if (!ref $hash) {
 	LOGCROAK("I was passed a non-hash!");
@@ -308,7 +320,7 @@ sub process_service {
     DEBUG "[DEBUG] processing service: $service";
 
     # Some fields that are nice to have in the plugin output
-    $hash->{'fields'} = join(' ', map {munin_get_node_name($_)} @$children);
+    $hash->{'fields'} = join(' ', @$field_order);
     $hash->{'plugin'} = $service;
     $hash->{'graph_title'} = get_full_service_name($hash);
     $hash->{'host'}  = $hostalias;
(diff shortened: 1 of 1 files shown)


steveschnepp:
That's a little bit too invasive for a bugfix. I'm more eager for such a change in devel than in stable-2.0.
Also, I'm rewriting the whole munin in devel to use SQL metadata. 

Better come over to IRC when you have big changes in mind.

driskell:
@steveschnepp That's understandable.

It's probably bigger than needs be as I left some of my debug messages in. If I remove them would it be suitable? It's just two changed blocks essentially - the work queue loop and the field order loop.

I'm also not sure on devel vs stable-2.0 - I've been asked previously to PR to stable-2.0 and that it would be merged to devel too? Or would I need two PR? My target is 2.0 but if it can be kept in devel that'll help future rewrite not repeat the mistake.

driskell:
What times for IRC? Might be able to help update this hit for you as I'm beginning to understand the Limits code quite well.

steveschnepp:
We have a *global* presence on IRC. But as we are not always reactive, you might have to wait several hours after your question to have an answer.

About not merging, it might get merged once I have a better understanding of the patch. It's just that it was "big" (which is good), and the LimitsOld code is older than my own contributions, so I'm not an expert in it, and touch it only seldomly, which very limited patchs.

steveschnepp:
... let's reopen the PR. Could you rework it to make it minimal ?

driskell:
I've crammed it down as far as I can, to the minimum required to fix the issues.
I looked at splitting the two issues (inheritance and aliased fields issues) into separate but they conflict.

It's likely still difficult to understand unless you're familiar so maybe I'll drop onto IRC at some point if I can find time and try to explain the issue the best I can, and the way I approached the fix, so we can review this better.

Thanks!

ssm:
The code and logic looks good.

I'd like to make a test for it, which should fail on the stable-2.0 branch, and pass with this pull request before merging.

(This would then be applied to the "devel" branch, to fix the problem there, too, if it is present, which it very well may be.)

grooverdan:
Thanks @driskell looking forward to the fix.

driskell:
@ssm Is this still on the radar? I can try help with the test if need be.

ssm:
I've no time to test this properly (or anything else for a while)

I've unassigned myself so as not to be a roadblock, in case anyone else would like to test and merge.  The code and logic still looks good.

driskell:
@ssm No worries.

If it helps, I've been using this patch in Production for several months now on at least 6 different setups, and my alerting has been flawless. I've not needed to touch anything since.

Specifically my cases that were broke that now work:

# Mongo plugin: `mongo_lag.warning` and `mongo_lag.critical` now work properly. The graph actually contains many fields, each field label is the hostname of the cluster member, and the value is the latency. Without this patch I need to hard-code in hostnames on every machine so it can find the field and apply the limit. With this patch, I can omit the field name and it inherits the limit from the graph level into all fields. Works great for me.
# JMX plugin: `jmx_threads.threads.Threads.warning` now works properly. Previously it would be ignored and never alert because there is a field alias involved. Now it works great.

driskell:
@ssm I've not looked at virtual nodes yet. But if it's improved it that's great. I can have a deeper look at fixing the value though.

Were you able to test aliased fields? My specific test case was the "jmx_" plugin, specifically "jmx_Multigraph" - this creates separate data graphs and visual graphs I think, and the visual graphs refer back to the data ones via aliases. Before this patch, setting warning/critical on the threads would never work - it would show inside the GUI for the graph where it lists the fields and thresholds, but it would never apply it to the graphs or trigger any alerts.

I've replicated the "diskstats_latency" result you provided. It turns out to be due to the removal of the following (which had to be removed to allow inheritance):

```
# Test directly here as get_limits is in truth recursive and
	# that fools us when processing multigraphs.
	next if (!defined($field->{warning}) and !defined($field->{critical}));
```

The following that was kept should have been enough:

```
# Skip fields without warning/critical definitions
        next if (!defined $warn and !defined $crit);
```

However, it wasn't enough - it actually turned out to be a NO-OP. `get_limits` returns `[undef, undef]` when there is no threshold. So, actually, `$crit` and `$warn` are always defined because the reference is defined! Thanks for spotting that! I've patched and will push up a change to the PR shortly. The patch modifies `get_limits` so that it returns an undefined reference if there is no threshold rather than a valid reference to an array.

driskell:
Hmmmm, I am wondering which way would be preferred.
We can enhance the check in `process_service` so it checks for `[undef, undef]` - or modify `get_limits` to return actual `undef`. I'm leaning towards the latter as in my opinion `undef` is clearer than `[undef, undef]` - but it's more lines changed in the PR. I'll push the latter. If it's decided it's better to only change the one line in `process_service` to a bigger and more complex check - let me know.

I've also verifies `get_limits` is not used anywhere else except `process_service`.

driskell:
The issue with the virtual node is that it refers to a value in a graph that belongs to a different server. The server the virtual node is for is "Overview".

The way the limits processing works out changes in values is by loading the values from a state file for the server. This is loaded once at the beginning of `process_service` so it only loads the values for the server the graph belongs to. Thus in this case - it loads an empty statefile for `localhost.localdomain;Overview` since the values are stored in the statefile for `localdomain;localhost.localdomain`.

It will be a big change I think to fix that as we'd need to work out which state file to load for every field using the new `$path` variable which is set for aliased fields - and out of scope I think for what this PR is intending to fix. It's outside of my requirements but if it's desired and I find some time I could take a stab at it in another PR. Fairly busy at the moment though so it wouldn't be in the coming few weeks.

driskell:
Just to add - the virtual node reporting is improved in this PR - but still broken - due to other issues still present in LimitsOld.

ssm:
Lovely.  I'll give it a test again shortly.

(I really must make myself an automated test bench for this)

driskell:
Hey @ssm - let me know if there's anything I can do to help with this.

PR #366, opened 2015-03-19
Feature/remove old relics

ssm:
Removes unused commands and their modules


steveschnepp:
:+1: removing that is indeed a feature

PR #372, opened 2015-03-21
Feature/more robust async

ssm:
Replace the constructors of Munin::Node::SpoolReader and Munin::Node::SpoolWriter. All data used is validated, although validation could be stricter.
 
Adds Munin::Common::Utils, with an is_valid_hostname function, used by async modules.


steveschnepp:
@ssm i just merged it, since otherwise it will get lost

PR #385, opened 2015-03-30
mysql: Add user statistics graphs

grooverdan:
Graphs connections, time, bytes, rows and commands. All broken down per
user.


grooverdan:
thank you :-)

PR #386, opened 2015-03-30
mysql: change default database to information_schema

grooverdan:
The previous default connected to the mysql database. This required
select privs on the mysql database to connect, however nothing in the mysql
database was actually used by the plugin.

It seems perl DBI insists on a database, or at least is more consistent
to specify one.

Using the information_schema this requires no additional priviledges
from the munin user. It has also been there since at least mysql-5.0.

While we are at it, the permissions for the munin user are documented
and a few more configuration variants are documented.

### DIFF
--- a/plugins/node.d/mysql_
+++ b/plugins/node.d/mysql_
@@ -57,6 +57,7 @@ Non-default example:
     env.mysqlpassword ryuWyawEv
     env.cachenamespace munin_mysql_alt
   [mysql10_*]
+    user munin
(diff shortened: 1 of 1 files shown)


grooverdan:
the information schema was in mysql-5.0 addition (released in 2005).

steveschnepp:
I'm loving those "morning" PR :+1:

grooverdan:
loving the quick "afternoon" merges too :+1:

PR #389, opened 2015-04-02
Fix several bugs in the RRD tuning

ssm:
Bugfix: Make Munin skip nonexistent RRD files when tuning.  Munin-update logged one error message per missing RRD file per new plugin/node when tuning.  This should make less noise in the logs.

Logging: Get the error message from RRD when tuning, and ensure it ends up in the log.

### DIFF
--- a/lib/Munin/Master/UpdateWorker.pm
+++ b/lib/Munin/Master/UpdateWorker.pm
@@ -554,23 +554,38 @@ sub _ensure_filename {
 
 
 sub _ensure_tuning {
-    my ($self, $service, $data_source, $ds_config) = @_;
+    my ( $self, $service, $data_source, $ds_config ) = @_;
+    my $fqn = sprintf( "%s:%s", $self->{ID}, $service );
+
     my $success = 1;
 
-    my $rrd_file =
-        $self->_get_rrd_file_name($service, $data_source,
-                                  $ds_config);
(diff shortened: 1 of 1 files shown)


lib/Munin/Master/UpdateWorker.pm
```diff
@@ -421,70 +421,75 @@ sub uw_override_with_conf {
...
+# Tune any RRD files where the metadata on disk differs from the
+# metadata received by munin-update.
 sub _compare_and_act_on_config_changes {
     my ($self, $nested_service_config) = @_;
 
-    # Kjellm: Why do we need to tune RRD files after upgrade?
-    # Shouldn't we create a upgrade script or something instead?
-    #
-    # janl: Upgrade script sucks.  This way it's inline in munin and
-    #  no need to remember anything or anything.
```
steveschnepp: I'd say that after a while in the trenches, I'm slowly in favor of upgrade scripts. Auto updating GAUGE/DERIVE seems a very bad idea after all. 

On the other side, updating the min/max looks nice. But I'm not sure that *ignoring* data is something that should look as innocent as it currently looks.

steveschnepp: ... what about using min/max for 
* auto-configuration of warnings and critical values
* showing the graphs

PR #391, opened 2015-04-02
Documentation update for 2.1.x

ssm:
Updates the install documentation for the devel branch


steveschnepp:
i'm asking myself about having a 2.0 version of it. maybe we should backport some changes there.

ssm:
Most of this is 2.1 stuff.  The one thing I think would make sense is to
replace the large INSTALL with a small INSTALL.rst.  Anything else?

steveschnepp:
*it* wasn't about this particular PR, but the whole doc.

PR #411, opened 2015-04-05
Enhance overview page

chteuchteu:
- Use SVG logo in header instead of PNG one (much better on high-res screens)
- Style nodes list
- Add a JS filter field to quickly find a node


steveschnepp:
hmmm, this needs a rebase to avoid extra merges

chteuchteu:
screenshot from 2015-04-05 20 35 56

Here's a screenshot of those enhancements for a quick overview

ssm:
Example node with no data to display, looks like the header

screenshot from 2015-04-05 21 46 08

ssm:
Here's the relevant HTML tags

    <li>
                    <span class="domain"><a href="munin-monitoring.org/">munin-monitoring.org</a> </span>
                    
                    
                    <ul>
                        
                            <li class="last">
                            <span class="domain"><a href="munin-monitoring.org/dho-nha.munin-monitoring.org/">dho-nha.munin-monitoring.org</a></span>
                            
                            
                            <ul>
                                
                            </ul>
                            </li>
                        
                    </ul>
                    </li>

chteuchteu:
Thank you for those information!
It seems that this node is considered as a domain (see the class="domain" its direct parent).

Is this a standard behavior or a bug in the way data is passed to templates? If not, I could restrict the CSS so it applies only to the top-level domain

ssm:
"Quentin S." <notifications@github.com> writes:

> It seems that this node is considered as a domain (see the
> class="domain" its direct parent).
>
> Is this a standard behavior or a bug in the way data is passed to
> templates? If not, I could restrict the CSS so it applies only to the
> top-level domain

That's a bug somewhere in the perl code.

The data passed to the template should report a host as a host, not a
domain.  I've added #413 to track that bug.

-- 
Stig Sandbeck Mathisen

ssm:
Rebased and merged manually as 09d57952d0bf6340cdf694caaeacdadcc7c2b400

PR #415, opened 2015-04-06
Fix bugs with time range switchs

chteuchteu:


ssm:
Rebased and merged as abca4f02263e619ad9dc50aa050968c304139d71

PR #423, opened 2015-04-07
Misc. UI improvements

chteuchteu:
- New sharp favicon. Please note that a lot of favicons are made available for each platform (when one pin the page on Windows 8+ / iOS / Android, creates a shortcut of the page, ...). That may be a detail, but it counts
- JS refresh (instead of reloading each page): reduce bandwidth consumption
- Store time range switch value in URL, and retrieve it on page load


chteuchteu:
(currently trying to figure out why the build failed)

PR #428, opened 2015-04-08
Topic carbon

darac:
I thought it might be interesting to look at exporting Munin statistics to the Carbon/Graphite graphing system. So, this is a fairly lightweight edit which exports statistics at the same time as it writes to RRD. The storage schema is a simple "`${prefix}.${reversed hostname}.${service}.${ds_name}`" (where prefix is optional and configurable).

I've not actually, played much with Carbon/Graphite yet, but this does at least show up on a test server.


steveschnepp:
Let's go crazy, and merge it.

PR #450, opened 2015-04-23
Remove all R_PATH usages, misc. UI improvements

chteuchteu:
According to what has been said in IRC meeting, I removed all usages of R_PATH var. We'll have to find a workaround for HTML files generated with munin-cron since this breaks resources includes and links for these files.

There also two bugfixes (see the commits)


szepeviktor:
@chteuchteu My munin website is in a subdir intentionally - for security - and after removing R_PATH resources are loaded from the root. Could you help me?

chteuchteu:
This is by design: using munin-httpd, pages will be served on another port (4948?). That'll be easier to make it cohabit with other web servers (we don't have to mess with apache configuration anymore)

Couldn't you use a proxy to server pages in a subdirectory instead of the root?

szepeviktor:
```apache
Alias /static /var/cache/munin/www/static
```

So new munin installs should use munin-httpd?

PR #462, opened 2015-04-30
First pass at fleshing out plugin documentation.

shevett:
Per the meeting on Freenode today, I was tasked with updating the plugin documentation to show guidelines and clarify error conditions and other helpful hints for writing plugins.  I also reworked some of the existing docs to make them more 'doc like' (not written in second person, etc).  The index page now has sub references to the secondary pages, and is a little more complete.


ssm:
Thanks :)

PR #472, opened 2015-05-05
Add binlogs*commits graphes to mysql plugin

roidelapluie:


ssm:
I've merged #473 instead of this, which provides the same monitoring, in addition to a few other data sources in the same graph.

Thanks for the pull request.

roidelapluie:
@ssm @grooverdan Nice, but can we consider using the COUNTER type? What do you think?

grooverdan:
86abbe2a5b2383b13b580b9c60ed68b5e139fdf4 was added to the docs very recently. DERIVE is the default type. The underlying structure is a unsigned long long which puts in the the won't overflow for many many years at high throughput category.

roidelapluie:
is min = 0 also the default?

PR #491, opened 2015-05-31
Linux plugins for wireless link quality, signal, and noise.

rvandegrift:


kenyon:
These would probably be better as a single multigraph plugin. Also, they should be submitted to the contrib repository first: https://github.com/munin-monitoring/contrib

sumpfralle:
In the meantime I added a plugin with a slightly broader scope to the contrib repository ("wifi_signal_noise_").
Thus this pull request can probably be closed.

PR #502, opened 2015-06-23
Fix buffer memory accounting

AMDmi3:
Despite what top(1) says, buffer memory is not a part of wired, and
it can be take more memory that wired. Thus do not subtract it from
wired and display as a separate line.


AMDmi3:
Here's what new graph look like, and you can see buffer memory > wired.

memory-day

AMDmi3:
Thank you!

ssm:
Likewise, thank you for the pull request. :)

PR #554, opened 2015-08-27
split cache into cache + shmem in linux memory plugin

rrzefox:
In its current form, the standard munin "memory" plugin on linux will generate very misleading graphs on many machines.
That is because it doesn't use the `Shmem` information from `/proc/meminfo`. All `shmem` will just be drawn as `cache`.
There is however a major difference between `shmem` (which includes the space used by tmpfs and SysV SHM segments) and filesystem cache: The FScache can be freed when the memory is needed for other things, the shmem cannot.

Therefore, the generated graphs are very misleading for machines where a lot of shmem is used, like certain database servers or servers using large tmpfs ramdisks: They suggest that the machine is not low on memory, because the memory is just used by `cache`, which one assumes to be freeable when needed, when in fact that memory is permanently bound in `shmem` and unfreeable.

This patch modifies the plugin to use `shmem` information (if available in `/proc/meminfo`).
Note that this patch tries to be minimal, really only adding the shmem info and doing nothing else - it even keeps the coding style and formatting of the original plugin. Much more could be improved on this plugin, as it seems it was last updated in early 2.6-kernel times, and since then a lot more useful information became available in `/proc/meminfo`.

Here is an example before / after pic:
munin-memory-day-after

### DIFF
--- a/plugins/node.d.linux/memory
+++ b/plugins/node.d.linux/memory
@@ -87,6 +87,8 @@ if ($ARGV[0] and $ARGV[0] eq "config") {
 
     print "slab " if exists $mems{'Slab'};
 
+    print "shmem " if exists $mems{'Shmem'};
+
     print "cached ",
       "buffers ",
 	"free ",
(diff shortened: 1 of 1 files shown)


ssm:
A nice improvement, thank you.

PR #573, opened 2015-10-04
Implement .sum

steveschnepp:
This is to implement the .sum in plugins.

The .stack will be dropped in 3.0


steveschnepp:
.sum works now. .stack is still unimplemented

Issue #583, opened 2015-10-25
Devices, such as ZFS datasets, with spaces in them, makes df misbehave


rkarlsba:
Seems if I use a ZFS dataset (aka filesystem) with a space in it, the df plugin does not seem to like this much. This is from output of df -P

  data/home/mathias         52428800        128     52428672        1% /data/home/mathias
  data/tmp/test asdf      1071483136          0   1071483136        0% /data/tmp/test asdf

So it seems line 134 in df

  my ($name, undef, undef, undef, $capacity, $mountpt) = split(/\s+/, $_, 6);

should be rewritten to allow this, but I'm not quite sure how. Any ideas?

df doesn't seem to be changed in 2.1 as compared to 2.0. 

roy

rkarlsba:
This patch should work.

rkarlsba:
This patch should work http://karlsbakk.net/df-zfs-naming.patch

PR #584, opened 2015-10-25
Fix for #583

rkarlsba:
Replaced a split() with a regex to allow device names with spaces where that's supported, such as with ZFS datasets.

This one 

if (/^([\w\/\s]+)\s{3}(\d+)\s+(\d+)\s+(\d+)\s+(\d+\%)\s+(.*)/)

should possibly be replaced with

if (/^([\w\/\s]+)\s   (\d{4,})\s+(\d+)\s+(\d+)\s+(\d+\%)\s+(.*)/)

But I found no way to change the change :P
roy

### DIFF
--- a/plugins/node.d.linux/df
+++ b/plugins/node.d.linux/df
@@ -128,10 +128,22 @@ my (@include_re, @exclude_re);
 
 sub split_and_skip {
   my ($line) = @_;
+  my ($name, $capacity, $mountpt);
 
   # Parse the output
   chomp($_);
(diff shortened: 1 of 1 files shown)


rkarlsba:
As for this

if (/^([\w\/\s]+)\s (\d{4,})\s+(\d+)\s+(\d+)\s+(\d+\%)\s+(.*)/)

Make it

if (/^([\w\/\s]+) {3}(\d{4,})\s+(\d+)\s+(\d+)\s+(\d+\%)\s+(.*)/)

PR #585, opened 2015-10-30
Add an explanation why COUNTER is harmful

ms7s:


ms7s:
The explanation is what I come up with after a quick google search as it was not immediately obvious to me why COUNTER is considered harmful. Please update it if you have a better explanation or a relevant link. @steveschnepp should probably know the reason precisely as he added the original note. Thank you.

PR #605, opened 2015-12-06
Small fixes to if_ and if_err_ plugins

kimheino:


kimheino:
I think the travis failure is incorrect / bug in testsuite. I still would like to see change this merged.

ssm:
Yes, the diff in the test failure seems to be two unix epoch timestamps
one second apart. Something is up with that test.

PR #611, opened 2015-12-28
Use highres (SVG) badges in README

splattael:
## Before

munin before

## After

munin after

Note both badges for Travis CI.


ssm:
Looks much better, thanks

splattael:
@ssm Wooow that merge was _fast_! Thank you :)

PR #631, opened 2016-01-17
Hidpi

steveschnepp:


virusman:
@steveschnepp How do I use this? I'm not sure how to pass `hidpi` to this code..

sumpfralle:
> How do I use this? I'm not sure how to pass hidpi to this code.

I think, you can append `x2` (or other numbers) to each image path (e.g. `month.pngx2`). This feels a bit unconventional :)

Please note, that this piece of code is only available in the `master` branch (to be release as munin 3.0).

PR #638, opened 2016-01-18
Plugins' command parameter: don't split on escaped whitespace.

shapirus:
This allows to pass arguments containing whitespaces as single strings.


shapirus:
Has anyone tested it but me?

steveschnepp:
merging will give testing.

PR #639, opened 2016-01-19
Dynazoom: don't hardcode graphs extension, closes #632

### LINKED ISSUE #632: 2.999.2: Dynazoom uses PNG links irrespective of the graph_ext parameter
in munin-dynazoom.js:
```
function refreshImg() {
	var urlPrefix = f_cgiurl_graph.val() + (f_cgiurl_graph.val() != '/' ? '/' : '');

	var url = urlPrefix + f_plugin_name.val()
		+ "-pinpoint=" + parseInt(f_start_epoch.val()) + "," + parseInt(f_stop_epoch.val())
		+ ".png"
		+ "?size_x=" + f_size_x.val()
		+ "&size_y=" + f_size_y.val();

	if (f_lower_limit.val())
		url += "&lower_limit=" + f_lower_limit.val();

	if (f_upper_limit.val())
		url += "&upper_limit=" + f_upper_limit.val();

	image.attr('src', url);

	return ((parseInt(f_stop_epoch.val()) - parseInt(f_start_epoch.val())) / parseInt(f_size_x.val()));
}
```

As we can see, the ".png" part is hardcoded whereas it should be chosen dynamically depending on what was the value of graph_ext parameter (graph_ext=svg in my case) originally passed to the server (or configured otherwise, such as in munin.conf, when it gets implemented). In other words, the zoomed graph's URL should be constructed using the same extension that the regular graphs in the page have (be it PNG or SVG).

This likely requires support on the server side in order for the JS part to be able to tell what image format was used originally, or perhaps we can get it from the "src" property of the image that is being clicked on in order to produce the zoomed graph.

chteuchteu:


chteuchteu:
This PR contains unwanted changes, let's create a new clean one.

Issue #648, opened 2016-01-22
2.999.2: procfs plugin, graph cpu: wrong presentation of certain fields


shapirus:
While testing munin on some physical boxes hosting a few VMs (kvm) under Proxmox virtualization software, I encountered a strange issue where the cpu counters user+system+idle+guest add up to more than 100% on the CPU graph, and what was most interesting is that it was the guest value that was over the ncpu*100% mark.

While I was unable to confirm that in any documentation, it appears that the Linux kernel counts all other counters already including the guest one. Watching system load using top confirmed that to me: all of the VM load was counted as user, which in amount was almost the same as the guest field value on the graph.

Please see the attached graphs. They were recorded on a single 4-core CPU box.
One of them is done using the original procfs plugin, another one with the guest* fields draw type changed to LINE. In my case it's quite clear that they (I can't confirm the guest_nice field, though, since it's always zero on my boxes) should really be of LINE type rather than AREASTACK.

![cpu-day-guest-areastack](https://cloud.githubusercontent.com/assets/725402/12501883/9ef3fd82-c090-11e5-8f1c-a1b4585b99cf.png)
![cpu-day-guest-line](https://cloud.githubusercontent.com/assets/725402/12501884/9f034102-c090-11e5-9301-8a40c0db1e76.png)

(nevermind the weird color change on the left size of the graphs, that was due to PR #647 )

I am looking for someone to confirm that this is reproducible (or behaves in a different way) with other virtualization systems. I have nothing other than Proxmox/kvm to play with at the moment.

The proposed PR to fix this issue is #649.

shapirus:
UPDATE:
The old pull request (#647) for this issue has been closed and another one has been created instead: #651.

New information regarding this was revealed: in fact, Linux counts the guest ticks in both guest and user fields, therefore to graph the own host's user ticks separately from the guest ticks, we need to subtract the latter from the former.

See function account_guest_time() in kernel/sched/cputime.c:

```
        /* Add guest time to cpustat. */
        if (task_nice(p) > 0) {
                cpustat[CPUTIME_NICE] += (__force u64) cputime;
                cpustat[CPUTIME_GUEST_NICE] += (__force u64) cputime;
        } else {
                cpustat[CPUTIME_USER] += (__force u64) cputime;
                cpustat[CPUTIME_GUEST] += (__force u64) cputime;
        }
```
The new pull request fixes that as well as establishes a correct graph_order with guest* fields being below idle.
[...]

PR #649, opened 2016-01-22
procfs plugin: use LINE draw type for guest* cpu fields

shapirus:
See issue #648


shapirus:
This will be fixed in a better way, I'm closing this PR.

PR #650, opened 2016-01-22
Fix timerange switch on category-view

chteuchteu:


chteuchteu:
Some checks failed, but it does not come from this pull-request (JS & CSS files cannot trigger CI failures)

PR #661, opened 2016-01-29
Support additional rrdtool draw modifiers such as dashes

shapirus:
This patch allows plugins to provide extra draw type modifiers supported by rrdtool, such as dashes.

Example:

```
fieldname.draw LINE3:dashes=5,5
```

### DIFF
--- a/lib/Munin/Master/Graph.pm
+++ b/lib/Munin/Master/Graph.pm
@@ -379,6 +379,14 @@ sub handle_request
 		$_color = $COLOUR[$field_number % $#COLOUR] unless defined $_color;
 		$_drawtype = "LINE" unless defined $_drawtype;
 
+		# Support extra modifiers like ":dashes=..."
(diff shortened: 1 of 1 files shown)


ssm:
This will create backwards incompatibility with older munin masters for that plugin.

We might need to flag a capability from the master to the node, and check for it in the plugin.

shapirus:
Yes I'm aware of that. Can you think of any easy/preferred way to anounce capabilities to plugins?

steveschnepp:
usual way is to have a `MUNIN_CAP_DRAWEXT`

shapirus:
Great. I didn't realize it was possible to announce master's capabilities to plugins. That's the way to go, then.

shapirus:
It was agreed that introducing new capabilities on such minor occasions isn't a good practice.
Replaced by #666.

PR #669, opened 2016-02-02
Signal error in http_loadtime

szepeviktor:
If any of the monitored pages goes non-200 exit with error.

### DIFF
--- a/plugins/node.d/http_loadtime
+++ b/plugins/node.d/http_loadtime
@@ -102,7 +102,7 @@ cd $TEMPO_DIR || exit 1
 for uri in $urls
 do
     loadtime=`$time_bin -f "%e" wget $wget_opt --header='Accept-Encoding: gzip,deflate' $uri 2>&1`
-    if [ "$?" != "0" ]; then
+    if [ "$?" == "0" ]; then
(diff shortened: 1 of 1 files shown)


plugins/node.d/http_loadtime
```diff
@@ -102,9 +102,13 @@ cd $TEMPO_DIR || exit 1
 for uri in $urls
 do
     loadtime=`$time_bin -f "%e" wget $wget_opt --header='Accept-Encoding: gzip,deflate' $uri 2>&1`
-
-    esc_uri="$(clean_fieldname "$uri")"
-    echo $esc_uri".value $loadtime"
+    if [ "$?" != "0" ]; then
```
ssm: This looks reversed. I would expect "time" or "wget" to return 0 on success in a shell (as opposed to perl).

    if [ "$?" = "0" ]; then
      # do good things
    else
      # warn about error
    fi

szepeviktor: Excuse me.
I'd like to rewrite the whole plugin to make it much safer.

szepeviktor:
Thank you.

PR #679, opened 2016-02-19
ui/navigation: Fix broken link id

Nodraak:
*Hi, I just updated to Ubuntu 14.04 and the new munin version I got seems to have quite a few bugs. Here is a first fix :)*

Concerned version: Munin 2.0.19-3 (seems to be still present in branch *devel*).


chteuchteu:
Merged, thanks for spotting the issue!

PR #715, opened 2016-07-12
Log to file

kjetilho:
This adds a dispatcher for logs other than syslog or stdout.


ssm:
This was rebased in cbae1f1 and merged in 264780c

PR #724, opened 2016-08-03
change default port to "undefined"

danbst:
change default port to "undefined", according to program behavior

I noticed, that my munin-node started on 20203 port. So it is wrong to assume 4949 port to be default.

PS: using munin 2.999.2


steveschnepp:
Actually, the port is 4949 per default when connecting. Which is what `munin.conf` is about.

PR #726, opened 2016-08-04
Fix a few FreeBSD specific plugins, and add one.

mat813:
The FreeBSD ports tree has a [few patches](https://github.com/freebsd/freebsd-ports/tree/master/sysutils/munin-common/files) for munin plugins, I've been carrying them for a long time, and it just occured to me that the patches for FreeBSD specific plugins could be upstreamed.

So, this fix:
- if_\* plugins for recent FreeBSD versions,
- memory in case there is more than one swap device

And it adds an if_packets_ plugin to count packets instead of bytes (depending on the workflow, packets can be much more interesting than bytes)


acheronflow:
mat, we also fixed perl unordered dict problem for the bind9_rndc patch, see https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=195689 I don't know if that was ever upstreamed.

plugins/node.d.freebsd/if_packets_
```diff
@@ -0,0 +1,110 @@
+#!@@GOODSH@@
```
steveschnepp: Please use `/bin/sh` instead now.

Issue #602, opened 2015-11-27
smart_ plugin


svirusxxx:
On gentoo systems witch kernels > 4.0 I have problem witch smart_ plugin.

```
$ munin-run --debug smart_sda
# Processing plugin configuration from /etc/munin/plugin-conf.d/munin-node
# Setting /rgid/ruid/ to /120/0/
# Setting /egid/euid/ to /120 120/0/
# Setting up environment
# About to run '/etc/munin/plugins/smart_sda'
smart_sda: plugins' UID: 0 / plugins' GID: 120
smart_sda: No S.M.A.R.T device name found in plugin's symlink!
```

but /dev/sda exist , and smartcl works perfect.
also I use ssd_life plugin that using smartclt, and don't have any problems.

dipohl:
Please increase verbosity by changing False to True in the plugins source:

  ## You may edit the following 3 variables
  # Increase verbosity (True/False)
  verbose=False

and tell resulting output of:

  # munin-run --debug smart_sda config
  # munin-run --debug smart_sda

What version of munin-node are you using?

svirusxxx:
it's output witch verbose=True
```
# munin-run --debug smart_sda config
# Processing plugin configuration from /etc/munin/plugin-conf.d/munin-node
# Setting /rgid/ruid/ to /244/0/
# Setting /egid/euid/ to /244 244 6/0/
# Setting up environment
# About to run '/etc/munin/plugins/smart_sda config'
smart_sda: plugins' UID: 0 / plugins' GID: 244
smart_sda: No S.M.A.R.T device name found in plugin's symlink!
```
```
# munin-run --debug smart_sda
# Processing plugin configuration from /etc/munin/plugin-conf.d/munin-node
# Setting /rgid/ruid/ to /244/0/
# Setting /egid/euid/ to /244 244 6/0/
# Setting up environment
# About to run '/etc/munin/plugins/smart_sda'
smart_sda: plugins' UID: 0 / plugins' GID: 244
smart_sda: No S.M.A.R.T device name found in plugin's symlink!
```

munin-2.0.19 (stable in gentoo)
OS: Gentoo
Kernel: 4.1.7

dipohl:
Is there something like SELinux enabled on your server?

svirusxxx:
No, this is standard kernel compiled myself + patches from gentoo

On 5 December 2015 at 21:54, dipohl <notifications@github.com> wrote:

> Is there something like SELinux enabled on your server?
>
[...]

PR #728, opened 2016-08-20
Update smart_ for Python3

InsanePrawn:
Fix string index in python 3 as mentioned in #602


InsanePrawn:
Note that this is probably the first in a series of steps of making this compatible with python3 (which i'll probably look into in the following days), I wasn't thinking straight when I opened the pull request this early. Don't drink and open source!

PR #738, opened 2016-09-29
Use Encode::Guess in munin-graph to check whether plugin output is UTF-8 or Latin-1

iwaim:
munin 2.x breaks UTF-8 string in plugin output.
- http://munin-monitoring.org/ticket/839


steveschnepp:
Beware about also adding a dependency to `Encode::Guess`

PR #739, opened 2016-09-29
Fix netstat "active connection openings" string

acheronflow:
Netstat fixed some typos in stats strings, which breaks Munin's netstat plugin's "active connections" metric:

https://sourceforge.net/p/net-tools/code/ci/614e15d6f4e32d094288414f6bf086d9761539f8/

This patch corrects that.

### DIFF
--- a/plugins/node.d.linux/netstat
+++ b/plugins/node.d.linux/netstat
@@ -102,7 +102,7 @@ if [ "$1" = "config" ]; then
 fi
 
 netstat -s | awk '
-/active connections ope/  { print "active.value " $1 }
+/active connection ope/  { print "active.value " $1 }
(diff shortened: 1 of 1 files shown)


sumpfralle:
Could you change the regular expression (and comment it) in a way that works with both old and new versions of netstat?

acheronflow:
@sumpfralle, done. While I'm not an awk wizard, I tested and the fix works with gentoo's and FreeBSD's awk.

sumpfralle:
@coveralls: I guess, the braces around the "s" are not necessary. But anyway: it is clear and easily understandable. I think, it is good.

@steveschnepp: just in case you are waiting for me to merge this - I do not have these permissions (which is fine). Please go ahead, if you feel like.

PR #752, opened 2016-11-14
s/IMGWEEKSUM/IMGYEARSUM/

bokutin:


chteuchteu:
`<!-- .sum graphs.  One of the least used features of munin? -->`

If after all this time you're the only one that spotted this issue, it may indeed be the least used feature of munin :-)

Thanks for your contribution

Issue #762, opened 2016-12-14
munin graph replace cdef fieldname incorrectly


AndCycle:
munin-2.0.25

I am in the middle of writing some plugin, then I found cdef doesn't get correct fieldname replaced

here is the info

plugin config
```
graph_title ZFS ARC Efficiency
graph_args --lower-limit 0 --upper-limit 100
graph_vlabel %
graph_info This graph shows the ARC Efficiency
graph_category test
hits.type DERIVE
hits.min 0
hits.graph no
misses.type DERIVE
misses.min 0
misses.graph no
hits_pct.type GAUGE
hits_pct.cdef hits,100,*,hits,misses,+,/
hits_pct.label Hits ratio
demand_data_hits.type DERIVE
demand_data_hits.min 0
demand_data_hits.graph no
demand_data_misses.type DERIVE
demand_data_misses.min 0
demand_data_misses.graph no
demand_data_hits_pct.type GAUGE
demand_data_hits_pct.cdef demand_data_hits,100,*,demand_data_hits,demand_data_misses,+,/
demand_data_hits_pct.label Demand Data Efficiency
demand_metadata_hits.type DERIVE
demand_metadata_hits.min 0
demand_metadata_hits.graph no
demand_metadata_misses.type DERIVE
demand_metadata_misses.min 0
demand_metadata_misses.graph no
demand_metadata_hits_pct.type GAUGE
demand_metadata_hits_pct.cdef demand_metadata_hits,100,*,demand_metadata_hits,demand_metadata_misses,+,/
demand_metadata_hits_pct.label Demand Metadata Efficiency
prefetch_data_hits.type DERIVE
prefetch_data_hits.min 0
prefetch_data_hits.graph no
prefetch_data_misses.type DERIVE
prefetch_data_misses.min 0
prefetch_data_misses.graph no
prefetch_data_hits_pct.type GAUGE
prefetch_data_hits_pct.cdef prefetch_data_hits,100,*,prefetch_data_hits,prefetch_data_misses,+,/
prefetch_data_hits_pct.label Prefetch Data Efficiency
prefetch_metadata_hits.type DERIVE
prefetch_metadata_hits.min 0
prefetch_metadata_hits.graph no
prefetch_metadata_misses.type DERIVE
prefetch_metadata_misses.min 0
prefetch_metadata_misses.graph no
prefetch_metadata_hits_pct.type GAUGE
[...]

PR #763, opened 2016-12-15
one line fix to correct replace fieldname in expand_cdef

AndCycle:
this is an attempt to fix https://github.com/munin-monitoring/munin/issues/762

but I never write any perl before, so this require review or rewritten in order not break anything.


AndCycle:
this is a nono one, at negative cdef also suffer same issue

PR #765, opened 2016-12-16
Fix RRD filename with subhosts

shirosaki:
Fix issues with ESX plugin.
http://munin-monitoring.org/ticket/1224

We assume that a host node has address.


kenyon:
Related: http://munin-monitoring.org/ticket/1236

sumpfralle:
Thank you for your proposal!

Meanwhile this part of the code was removed due to the move to a database state storage. Thus the problem should not exist anymore. Thank you anyway!

PR #774, opened 2017-01-10
Remove --sdtout param from munin-update command

Lusitaniae:
I was going through the troubleshooting documentation, and one of the commands suggested is using an outdated(?) parameter that doesn't exist.

Error:

    $ /usr/share/munin/munin-update --debug --nofork --stdout --host example.com --service df

    Unknown option: stdout

Version installed: 

    $ /usr/share/munin/munin-update --version                                       

    munin version 2.0.25-2.

OS installed:

    cat /etc/centos-release
 
    CentOS release 6.8 (Final)

Disclaimer: This is the first time going through Munin in detail.

### DIFF
--- a/doc/tutorial/troubleshooting.rst
+++ b/doc/tutorial/troubleshooting.rst
@@ -205,7 +205,7 @@ Run :ref:`munin-update` as user ``munin`` on the Munin master machine.
 ::
 
   # su -s /bin/bash munin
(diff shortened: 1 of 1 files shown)


kenyon:
Fixed in commit 3730cf37acdb23b70e11e85e6ef12470ef9ebd68. Thanks.

PR #785, opened 2017-01-17
Various commits to handle 1k SNMP nodes

steveschnepp:
This is a PR for cherry picking useful commits in order to achieve 1k SNMP nodes.


sumpfralle:
What do you: would these changes be good additions for the 3.0 branch?

PR #796, opened 2017-02-22
snmp__df_ram: allow individual critical and warning levels

bruderjakob12:
To my knowledge it was not possible to specify warning and critical levels. With this patch the levels are read from the configuration. It is possible to change the values for all measurements with env.warning/env.critical and for each measurement with env.{measurement}_warning/env.{measurement}_critical.

### DIFF
--- a/plugins/node.d/snmp__df_ram
+++ b/plugins/node.d/snmp__df_ram
@@ -51,6 +51,8 @@ my $MAXLABEL = 20;
 my $host      = $ENV{host}      || undef;
 my $port      = $ENV{port}      || 161;
 my $community = $ENV{community} || "public";
+my $warnlevel = $ENV{warning}   || "92";
+my $critlevel = $ENV{critical}  || "95";
 
 my $response;
 
@@ -178,12 +180,26 @@ if (defined $ARGV[0] and $ARGV[0] eq "config")
 
 	foreach my $part (keys %partitions)
 	{
-		print (&get_name_by_mp ($part), ".label ");
-		print (length($part)<=$MAXLABEL ? $part : "...".substr($part,-($MAXLABEL-3)));
-		print ("\n");
-		print (&get_name_by_mp ($part), ".warning 92\n");
-		print (&get_name_by_mp ($part), ".critical 98\n");
-		print (&get_name_by_mp ($part), ".info Usage for ". ($partitions{$part}{extinfo}||$part)."\n");
+                my $unit = &get_name_by_mp ($part);
+                print ($unit, ".label ");
+                print (length($part)<=$MAXLABEL ? $part : "...".substr($part,-($MAXLABEL-3)));
+                print ("\n");
+                if (defined $ENV{join('', $unit, "_warning")})
+                {
+                        print ($unit, ".warning ", $ENV{join('', $unit, "_warning")}, "\n");
+                } else
+                {
+                        print ($unit, ".warning ", $warnlevel, "\n");
+                }
+                if (defined $ENV{join('', $unit, "_critical")})
+                {
+                        print ($unit, ".critical ", $ENV{join('', $unit, "_critical")}, "\n");
+                } else
+                {
+                        print ($unit, ".critical ", $critlevel, "\n");
+                }
+                print ($unit, ".info Usage for ". ($partitions{$part}{extinfo}||$part)."\n");
+
 	}
 	exit 0;
 }


dipohl:
bruderjakob12 wrote
> To my knowledge it was not possible to specify warning and critical levels. 

You can always specify the warning and critical level (on global level for the plugin and also on level datasource, see below) in the nodes plugin configuration (/etc/munin/plugin-conf.d/*). For this purpose it doesn't need to be declared within the plugin. e.g. see 

https://github.com/munin-monitoring/munin/blob/stable-2.0/plugins/node.d.linux/cpu.in

`------------------ snip ---------------
EXAMPLE WARNING AND CRITICAL SETTINGS

You can also set warning and critical levels for each of the data
series the plugin reports.  The following environment variables are
used as default for all fields:

  env.warning
  env.critical

But each field can be controlled separately:

  env.system_warning
  env.system_critical

  env.user_warning
  env.user_critical

  env.nice_warning
  env.nice_critical

  env.idle_warning
  env.idle_critical
------------------ snip ---------------`

bruderjakob12:
Hi diphol, I tried to change the warning and critical values of the snmp__df_ram this way, but it had no effect. The warning and critical values for this specific mod are hardcoded in line 184 & 185 and directly printed so I don't see how the values could be changed afterwards.
A less quick and dirty way to patch the plugin would probably involve the use of Munin::Plugin::SNMP->config_session()

dipohl:
Hi bruderjakob12,

you wrote:
> I tried to change the warning and critical values of the snmp__df_ram this way, but it had no effect. 

Can you please share your configuration sections from these tries?

> The warning and critical values for this specific mod are hardcoded in line 184 & 185 and 
> directly printed so I don't see how the values could be changed afterwards.

You can see the _magic_ on the node with `"munin-run --debug <service-name> config"`.
The node overwrites directives given in the plugins output with the settings in files under `plugin-conf.d`.

I am not sure if this works for snmp plugins, though. Because these belong to another (virtual) node and not to the localhost node. I want to test this therefore see my wish above :)

**In anycase** it should be possible to overwrite warning and criticals for the (virtual) node in the Munin Master configuration.

http://guide.munin-monitoring.org/en/latest/tutorial/alert.html#syntax-of-warning-and-critical

fyi & cheers,

Gabriele

sumpfralle:
This is a bit old, but let us try to bring it to a good end!

> A less quick and dirty way to patch the plugin would probably involve the use of Munin::Plugin::SNMP->config_session()

This sounds like you have a better approach in mind. Since you revealed this knowledge, we will hardly let you slip through with the dirty hack :)
What specifically did you have in mind here?

Regarding the hardcoded values for warning/critical thresholds: instead you could use `print_thresholds($fieldname);`. This should do the proper thing in most situations.

> I shouldn't have to touch the master configuration.

The master is the place, where limits are processed and alarms are handled. Thus this is the proper place for overriding thresholds.

What do you think? How should we proceed?

bruderjakob12:
Unfortunately it seems that I misunderstood the purpose of `Munin::Plugin::SNMP->config_session()` so the usage of the environment variables seems to be the way to go.

plugins/node.d/snmp__df_ram
```diff
@@ -178,12 +180,26 @@ if (defined $ARGV[0] and $ARGV[0] eq "config")
...
-		print (length($part)<=$MAXLABEL ? $part : "...".substr($part,-($MAXLABEL-3)));
-		print ("\n");
-		print (&get_name_by_mp ($part), ".warning 92\n");
-		print (&get_name_by_mp ($part), ".critical 98\n");
-		print (&get_name_by_mp ($part), ".info Usage for ". ($partitions{$part}{extinfo}||$part)."\n");
+                my $unit = &get_name_by_mp ($part);
+                print ($unit, ".label ");
+                print (length($part)<=$MAXLABEL ? $part : "...".substr($part,-($MAXLABEL-3)));
+                print ("\n");
+                if (defined $ENV{join('', $unit, "_warning")})
```
sumpfralle: I think, I am not an obsessively terse code writer, but maybe the following could be comparably readable, but less confusing? (at least for my reading habits)

```suggestion
                print($unit, ".warning ",  defined($ENV{"${unit}_warning"}) ? $ENV{"${unit}_warning"} : $warnlevel);
                print($unit, ".critical ",  defined($ENV{"${unit}_critical"}) ? $ENV{"${unit}_critical"} : $critlevel);
```

(at least I would like to get rid of the `join`. Additionally I enjoyed to remove he duplicate `print`, too.
What do you think?

sumpfralle: Ping?

sumpfralle:
Please re-open, if you are still interested.

PR #801, opened 2017-03-01
[apt_all] Make @releases configurable

shtrom:
This avoids error messages when trying to update releases not present in
the sources. See Debian bug #715141. [0]

[0] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=715141


sumpfralle:
Thank you!

PR #808, opened 2017-03-13
Add some basic SMART sas monitoring

hamishcoleman:
With my (admittedly old) SAS drives, smartctl outputs a completely different format of data.  This adds basic scraping for that

### DIFF
--- a/plugins/node.d/smart_
+++ b/plugins/node.d/smart_
@@ -145,7 +145,7 @@ def read_values(hard_drive):
                 fields=string.split(l)
                 if 'temperature' not in smart_values:
                         smart_values['temperature']={}
-                smart_values['temperature']['threshold']=fields[3]
+                smart_values['temperature']['threshold_max']=fields[3]
         exit_status=smart_output.close()
         if exit_status!=None :
             # smartctl exit code is a bitmask, check man page.
@@ -219,7 +219,15 @@ def print_config(hard_drive) :
     for key in attributes :
         if key in ['smartctl_exit_status','model'] : continue
         print(key+'.label '+key)
-        if report_warnings: print(key+'.critical '+smart_values_state[key]["threshold"]+':')
+        if report_warnings:
+            threshold_min=''
+            threshold_max=''
+            if 'threshold' in smart_values_state[key]:
+                threshold_min=smart_values_state[key]["threshold"]
+            if 'threshold_max' in smart_values_state[key]:
+                threshold_max=smart_values_state[key]["threshold_max"]
+            if threshold_min or threshold_max:
+                print(key+'.critical '+threshold_min+':'+threshold_max)
     print('smartctl_exit_status.label smartctl_exit_status')
     if report_warnings: print('smartctl_exit_status.warning '+smart_values_state['smartctl_exit_status']["threshold"])
 
--- a/plugins/node.d/smart_
+++ b/plugins/node.d/smart_
@@ -134,6 +134,18 @@ def read_values(hard_drive):
             elif l[:18]=="ID# ATTRIBUTE_NAME" :
                 # Start reading the Attributes block
                 read_values=1
+            # smartctl has a different output format for (my?) SAS drives.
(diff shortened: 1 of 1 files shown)


dipohl:
There are different types of devices, with different sort of smartctl report. See section [SMART READ AND DISPLAY DATA OPTIONS](https://www.smartmontools.org/browser/trunk/smartmontools/smartctl.8.in#lbAE) of the smartctl man page.

Plugin smart_ is intended  to work with ATA reports and plots the attributes _normalized_ values (not the _raw_ values as you wrote in your comment in the patch)

Have a look at an [smartctl ATA report](https://www.smartmontools.org/wiki/Howto_ReadSmartctlReports_ATA_542.1). You see there the column "Value" (Mouse over brings shows you detail explanation in a pop-up window) which contains the _normalized_ value of the attributes. 

I would not like to see and use a smart_ plugin which present a mix-up of both sorts of values.

For temperature monitoring of SAS disks you can use the plugin [hddtemp_smartctl](https://github.com/munin-monitoring/munin/blob/master/plugins/node.d/hddtemp_smartctl) which is based on smartctl scsi report.

hamishcoleman:
There are a number of other SAS/SCSI only counters that smartctl can show - the temperature was simply the first one on the page as I worked my way down.  Just on the drives I have to test with, there are start/stop cycles, defect list size, blocks handled, error counts, power on hours, background scan data.  I figured that since smart_ is generating one graph for each drive and that each drive is either generating ATA-style attributes or SCSI-style attributes then the two types of values would not become too mixed together.

I was playing fast and loose with my use of "raw" and "cooked" too - what I should have said was that the sue of Centigrade as the measurement value means that the bounds checking was a upper threshold, instead of the usual ATA attribute checking which moves toward zero and thus is a lower threshold, and so additional threshold handling needed to be added.

I've had a look at the hddtemp_smartctl (thanks for reminding me about that - I had forgotten it) and it doesnt actually read or handle the drive specified temperature thresholds (and, obviously, it only handles temperature metrics)

Should I just write an entirely new plugin for SCSI/SAS style metric data?

dipohl:
> Should I just write an entirely new plugin for SCSI/SAS style metric data?

That would be fine! :-)

It may make sense to develop a wildcard plugin `smartscsi_<type>_`
with type = 
* temp (? not sure, because we have a plugin for this already..)
* cycle
* blocks
* errors
* ..

to set the appropriate scale of the graph for each sensor.

And with that I (as a user of the new plugin) would like to combine the data for all my disks on the system in one graph. That could be done with environment value _drives_ like in plugin [hddtemp_smartctl](http://gallery.munin-monitoring.org/distro/svn/munin-stable-2.0/plugins/node.d/hddtemp_smartctl.html)

btw: to support your work on the plugin with a review of the range of output variants in SCSI/SAS smartctl reports I posted a [request for contribution of smartctl reports for recent smartctl reports](https://sourceforge.net/p/smartmontools/mailman/message/35722268/) on smartmontools user list

Will you also contribute your reports (using a recent smartctl version) to our (smartmontools) wiki? I will be glad about it!

Thanks for your offer to develop the new plugin!

hamishcoleman:
unfortunately, I cannot add anything to the smartmontools wiki, because I do not exist:

> Submission rejected as potential spam
>
>    BotScout says this is spam (Y|MULTI|IP|0|MAIL|0|NAME|1)

hamishcoleman:
I will force a different set of commits into this PR when I have a minimally functional plugin.

sumpfralle:
Any news for this topic?

hamishcoleman:
sorry @sumpfralle , no - I built a prototype but thought I was overengineering it, then I got pulled onto other tasks

sumpfralle:
Ping?

hamishcoleman:
hey @sumpfralle, I no longer support any SAS drives - they have been either NVME or SATA for a while now, so the proposed SAS support is not something that I can test, let alone get allocated time to work on

PR #812, opened 2017-03-15
Fix solaris io_* plugins

shakemid:
Hello,
I would like to make a proposition to fix for Solaris io_busy_, io_bytes_, io_ops_ modules.
I have tested them on Solaris 11.3 and they look working fine. Please consider this pull request.


shakemid:
In my opinion, those plugins look less readable codes like bad example in http://munin-monitoring.org/wiki/PluginShell .
If I may, I would like to rewrite them.

shakemid:
In addition, I also wrote new io_disk plugin https://github.com/munin-monitoring/contrib/pull/826
It can replace io_busy_, io_ops_, io_bytes_ plugins and can take over RRDs generated by them. Would you consider new plugin to replace them?

PR #813, opened 2017-03-15
Tomcat improvements

bldewolf:
Two fixes for the perl tomcat plugins.  One fixes it so the username/password to tomcat can have URI unsafe characters (which tend to appear in passwords).

The other is backwards incompatible because it changes the datafield names so that it can support collecting data from more than one collector (otherwise we can only monitor a single collector per plugin instance), as well as auto-detecting collector names so that the plugins "just work".  They should still function in the mode where the admin specifies the collector name to support existing setups.

I also have a patchset for 2.0 available if these changes sound like a good idea for that branch too.

### DIFF
--- a/plugins/node.d/tomcat_access
+++ b/plugins/node.d/tomcat_access
@@ -54,6 +54,7 @@ use strict;
 use warnings;
 
 use Munin::Plugin::HTTP;
+use URI::Escape;
 
 my $ret = undef;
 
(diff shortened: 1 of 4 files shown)


sumpfralle:
Thank you for improving these plugins!

> I also have a patchset for 2.0 available if these changes sound like a good idea for that branch too.

I think we should not apply this change of the fieldname to the stable-2.0 branch, since it only receives maintenance releases.

PR #816, opened 2017-03-22
Add support for rndc options in bind9_rndc plugin.

sebastic:
The `bind9_rndc` plugin doesn't support options to set a different key (`-k`), config file (`-c`) or any of the other [`rndc` options](https://ftp.isc.org/isc/bind9/cur/9.11/doc/arm/man.rndc.html).

I needed these changes to monitor multiple BIND instances.

### DIFF
--- a/plugins/node.d/bind9_rndc
+++ b/plugins/node.d/bind9_rndc
@@ -12,6 +12,7 @@ The following environment variables are used by this plugin
 
   [bind9_rndc]
     env.rndc		/usr/sbin/rndc
+    env.rndc_options	
     env.querystats      /var/run/named.stats
 
 The user/group that runs the plugin must have read access to the stats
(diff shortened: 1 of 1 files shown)


ssm:
Have you tried setting env.rndc in your munin plugin configuration file to '/usr/bin/rndc -k key -c /path/to/config'?

sebastic:
> Have you tried setting env.rndc in your munin plugin configuration file to '/usr/bin/rndc -k key -c /path/to/config'?

No. While that may work, this is a cleaner solution, separating the command from its (optional) arguments.

ssm:
It allows for a cleaner solution in the future, I agree with that. :)

(The current plugin will treat the two variables a bit like http://dilbert.com/strip/1997-08-16)

PR #825, opened 2017-04-22
Fix some plugin categories and wired squeezebox devices

kimheino:


plugins/node.d/squeezebox_
```diff
@@ -192,13 +192,14 @@ elif [ "$CMD" = "signalstrength" -o "$CMD" = "mixer volume" ]; then
             MAC=$(printf "%b" "player id $ID ?\nexit\n" | $NC $HOST $PORT | cut -d " " -f 4 | sed 's/%3A/:/g')
             VAL=$(printf "%b" "$MAC $CMD ?\nexit\n"| $NC $HOST $PORT | cut -d " " -f 2- | sed "s/$CMD //")
             MAC2=$(echo $MAC| sed 's/://g')
+            [ $VAL -eq 0 ] && VAL=100
```
dipohl: This will set VAL to 100 for every VAL that has value 0.
Is it possible to recognize the wired devices which you want to address with this change?

kimheino: Sorry, my commit message was too short. 

This API returns wireless signal strength in percent. There is no "wired or wireless?" API query.

- wired devices are reported as 0
- connected wireless devices are reported as 1-100, with typical values as 80-100
- disconnected wireless devices are missing from the list, or reported as "last seen value"

dipohl:
Thanks for clarifying the signalstrength change!

PR #834, opened 2017-05-15
Fixing zoom

Night1:
Zoom required additional rewrites in order to function on nginx


Night1:
Thanks  to be0rn on irc for help in debugging.

PR #837, opened 2017-05-19
Fix alerts for unknown values not being triggered for GAUGE values

driskell:
Currently, the logic in the Limits process means it only checks for "out-of-date" values if the value is a COUNTER or DERIVE type. For GAUGE types it just processes the last received value.

This means if you shutdown a server running munin-node with disk usage and if_err plugins enabled with limits set on them, the Limits process on the munin master will only report unknown for the if_err plugin (which is COUNTER I think.) There will be no alert for disk usage.

After applying this PR, you receive alerts for unknown on both disk usage and if_err.

### DIFF
--- a/master/lib/Munin/Master/LimitsOld.pm
+++ b/master/lib/Munin/Master/LimitsOld.pm
@@ -369,13 +369,17 @@ sub process_service {
 		my ($previous_updated_timestamp, $previous_updated_value) = @{ $state->{value}{"$rrd_filename:42"}{previous} || [ ] };
 
 		my $heartbeat = 600; # XXX - $heartbeat is a fixed 10 min (2 runs of 5 min).
-		if (! $field->{type} || $field->{type} eq "GAUGE") {
-			$value = $current_updated_value;
-		} elsif (! defined $current_updated_value || ! defined $previous_updated_value || $current_updated_timestamp == $previous_updated_timestamp) {
-			# No derive computing possible. Report unknown.
+		if (! defined $current_updated_value) {
+			# No value yet. Report unknown.
 			$value = "U";
 		} elsif (time > $current_updated_timestamp + $heartbeat) {
-			# Current value is too old. Report unknown. 
+			# Current value is too old. Report unknown.
+			$value = "U";
+		} elsif (! $field->{type} || $field->{type} eq "GAUGE") {
+			# Non-compute up-to-date value.
+			$value = $current_updated_value;
+		} elsif (! defined $previous_updated_value || $current_updated_timestamp == $previous_updated_timestamp) {
+			# No derive computing possible. Report unknown.
 			$value = "U";
 		} elsif ($current_updated_timestamp > $previous_updated_timestamp + $heartbeat) {
 			# Old value is too old. Report unknown. 
--- a/master/lib/Munin/Master/LimitsOld.pm
+++ b/master/lib/Munin/Master/LimitsOld.pm
@@ -369,7 +369,7 @@ sub process_service {
 		my ($previous_updated_timestamp, $previous_updated_value) = @{ $state->{value}{"$rrd_filename:42"}{previous} || [ ] };
 
 		my $heartbeat = 600; # XXX - $heartbeat is a fixed 10 min (2 runs of 5 min).
-		if (! defined $current_updated_value) {
+		if (! defined $current_updated_value || $current_updated_value eq "U") {
 			# No value yet. Report unknown.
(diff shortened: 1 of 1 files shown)


driskell:
@ssm Hi do you need anything from me to progress this?

driskell:
Hi @ssm @sumpfralle is there anything needed here? Is it still viable? Thanks

sumpfralle:
Thank you for contributing and for being persistent! :)

I looked through the rather intertwined piece of code that you tackled.
Please correct me, if my following assumptions are not correct.

The previous logic seemed to have followed this flow:
1. GAUGE type: use current value (if undefined: this will turn into `U` later on)
2. COUNTER/DERIVE/ABSOLUTE: use `U` if the current or the last value is undefined or if there was no update
3. COUNTER/DERIVE/ABSOLUTE: use `U` if current value is outdated
4. COUNTER/DERIVE/ABSOLUTE: use `U` if previous value is too old compared to the current value
5. COUNTER: use `U` if current value is below previous value
6. COUNTER/DERIVE/ABSOLUTE: use `U` if current value is undefined (`U`)
7. ABSOLUTE: calculate new change / time
8. COUNTER/DERIVE: use `U` if previous value is undefined (`U`)
9. COUNTER/DERIVE: calculate new change / time

Your commit would result in the following flow:
1. GAUGE/COUNTER/DERIVE/ABSOLUTE: use `U` if current value is undefined
2. GAUGE/COUNTER/DERIVE/ABSOLUTE: use `U` if current value is outdated
3. GAUGE: use current value
4. COUNTER/DERIVE/ABSOLUTE: use `U` if previous value is undefined or there was no update
5. COUNTER/DERIVE/ABSOLUTE: use `U` if previous value is too old compared to the current value
6. COUNTER: use `U` if current value is below previous value
7. COUNTER/DERIVE/ABSOLUTE: use `U` if current value is undefined (`U`)
8. ABSOLUTE: calculate new change / time
9. COUNTER/DERIVE: use `U` if previous value is undefined (`U`)
10. COUNTER/DERIVE: calculate new change / time

Thus your commit fixes the previously missing check for "outdated current value" of GAUGE values (just as you described it).

I think, the following additional changes would be helpful:
* swap 6 and 7: otherwise the comparison in 6 could fail with the current value being `U` (only checked later in 7)
    * this problem existed both before and after your commit
* move 9 before 6 (here again: the comparison in 6 could fail)
* remove `!defined $value ||` from the `De-taint` test below this block
    * your commit removed the possibility of an undefined `$value`
* bonus beautification: unify all sanity checks before the tests:
    * `my $is_current_valid = defined $current_value && ($current_value ne "U") && (time <= $current_updated_timestamp + $heartbeat);`
    * `my $is_change_valid = $is_current_valid && defined $previous_value && ($previous_value ne "U") && ($current_updated_timestamp <= $previous_updated_timestamp + $heartbeat) && ($current_updated_timestamp != $previous_updated_timestamp);`
    * this should simplify the intertwined flow of this small code monster quite a bit

What do you think?
Are you still in the mood to further improve this piece of code?

Otherwise I could also merge your commit as it is right now, since it is a good improvement, anyway.

driskell:
@sumpfralle Absolutely happy to look at those. I was hesitant at the start at doing too much change (I know everyone's likely volunteer and limited time to check things intensively!), but if you're happy I'll try and find some time soon to get those recommendations looked at and pass some more eyes over 👍  :smile:

Thanks!

sumpfralle:
> Absolutely happy to look at those.

Perfect!

btw: I enjoy doing reviews - please feel free to exploit this habit of mine :)

driskell:
@sumpfralle thanks I've completed it now!

I did not perform the additional beautifying as I felt it would create some long lines with multiple checks and I didn't feel that was any clearer. In fact by moving a few of the logic about, such as checking for unknown at same time as whether defined, it reduced number of branches, and in my opinion makes it very easy to follow by reading through each comment.

So we handle in this order:

1. Check for a valid current value, returning Unknown if invalid
1. Check for stale value, returning Unknown if stale
1. If we're non-computed then return the current value
1. Check for a valid previous value, returning Unknown if invalid
1. Check for stale value or no new value, returning Unknown if stale or no new value
1. If we're ABSOLUTE return the required calculation
1. Check if the COUNTER/DERIVE value has reset, returning Unknown if it reset
1. Return the required COUNTER/DERIVE calculation

Hope that works!

sumpfralle:
That looks good - thank you!

PR #843, opened 2017-06-01
plugins/postgres: Fix connection stats for PG 9.2-9.4

nichivo:
Commit dce46027a5fdb7d5cd44dc8cdb70ebcc425a949f introduced a query for
PG 9.4 but it counts all connections (including idle) as active.

This patch fixes the graph as described in the comments for https://github.com/munin-monitoring/munin/commit/dce46027a5fdb7d5cd44dc8cdb70ebcc425a949f

By removing the 9.4 query, the 9.5 query will be applied to PG 9.2-9.4. This has been successfully tested on live servers running PG 9.3 and on official Postgres Docker images from 9.2 to 9.4.


sumpfralle:
Any objections? Otherwise I would merge this ...

PR #845, opened 2017-06-03
switch to python3

cgzones:
update plugins `ipmi_sensor_` and `smart_`

### DIFF
--- a/Makefile.config
+++ b/Makefile.config
@@ -68,7 +68,7 @@ STATEDIR   = $(DESTDIR)/var/run/munin
 PERL       := $(shell which perl)
 
 # The python interpreter to use (used by some plugins)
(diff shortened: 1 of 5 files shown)


sumpfralle:
I like all your style changes - they turn the python2-only plugins into python2/3 compatible ones. It would be a good step forward to merge these.

Regarding the switch from python2 to python3 (your changes outside of the `plugins` directory): what effect should these changes have? Do they serve a specific purpose?

ssm:
@sumpfralle the changes outside the plugins directory ensures the new plugins are tested correctly, as far as I can see.

Got to test the smart_ plugin, and that seemed to work.  Python 3 support is nice. :+1:

PR #861, opened 2017-07-19
DERIVE does not need a max to avoid false spikes.

aowi:
100k interrupts/s is not unlikely on a busy system.

This mirrors the change done in 5497c8b3 for master.


sumpfralle:
Thank you!

Issue #721, opened 2016-07-27
munin-cgi-graph CGI::param security problem


sstj:
Running munin-2.0.25 on Gentoo. I observed this message in the logs

2016/07/26 21:57:54 [PERL WARNING] CGI::param called in list context from /usr/libexec/munin/cgi/munin-cgi-graph line 450, this can lead to vulnerabilities. See the warning in "Fetching the value or values of a single named parameter" at /usr/lib64/perl5/vendor_perl/5.20.2/CGI.pm line 404.

This allows injecting options into munin-cgi-graph (similar to http://munin-monitoring.org/ticket/1238 ), by doing something like this:

&upper_limit=500&upper_limit=--output-file&upper_limit=/tmp/test.txt 

which wrote the graph to /tmp/test.txt

h01ger:
https://bugs.debian.org/855705 has a patch for this issue.

h01ger:
Does this affect 2.0.6 too? And 2.999.6? (Enotime to check right now, but it would be good if someone did…)

carnil:
A CVE has been requested via https://cveform.mitre.org/

carnil:
This has been assigned CVE-2017-6188 by MITRE

mejo-:
I just gave 2.0.6 (from Debian/Wheezy) a try and indeed it's vulnerable too.

The proposed patch by Tomaž Šolc from [Debian Bugreport #855705](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=855705#5) fixes this particular vulnerability.

h01ger:
control: found -1 2.0.6-4+deb7u2
control: tags -1 pending
thanks

On Fri, Feb 24, 2017 at 01:37:55AM -0800, mejo- wrote:
> I just gave 2.0.6 (from Debian/Wheezy) a try and indeed it's vulnerable too.
> The proposed patch by Tomaž Šolc from [Debian Bugreport #855705](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=855705#5) fixes this particular vulnerability.

thanks, mejo, for confirming this both!


-- 
cheers,
	Holger

sumpfralle:
This issue is fixed, or?
* `stable-2.0`: 42ce18f24d3eae8be33526a198bf21e4f2330230
* `master`: the relevant code was replaced before (according to Steve Schnepp in [Debian #855705](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=855705))

PR #862, opened 2017-07-20
Do not send empty upper and lower limit from dynazoom

pes-soft:
JavaScript function refreshImg() in dynazoom.html sets upper_limit and lower_limit parameters in the image query. When there are no limits set ( during initial graph display for example ), refreshImg() sets parameters with empty values ( upper_limit=&lower_limit= ). Because of the recent fixes in processing of CGI parameters in munin-cgi-graph ( #721 ) empty values are accepted and passed to RRD, which fails with error: `Cannot convert '' to float`.

This pull request fixes the refreshImg(), so limits are used in query only if their value is present ( using ternary condition operator ).

### DIFF
--- a/master/static/dynazoom.html
+++ b/master/static/dynazoom.html
@@ -100,8 +100,8 @@ function refreshImg() {
 		+ "-pinpoint=" + parseInt(form.start_epoch.value) + "," + parseInt(form.stop_epoch.value)
 		+ ".png"
 		+ "?" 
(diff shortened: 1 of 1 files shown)


pes-soft:
Ooops. This is actually fixed already by using Scalar::Util::looks_like_number in munin-cgi-graph. My tests were run against distribution where this is not merged yet.

sumpfralle:
Good, that it is already fixed.
Nevertheless: thank you for caring!

PR #879, opened 2017-10-14
Change solaris plugin directory name 

shakemid:
Hello,
It looks good to change solaris plugin directory name, because perl $^O returns 'solaris' on Solaris.
However, uname command returns 'SunOS'. This change is not necessary for Munin 2.0.x


sumpfralle:
I am not sure, what was the base of the naming convention for the distribution-specific plugin directories:
```
node.d.aix
node.d.cygwin
node.d.darwin
node.d.debug
node.d.freebsd
node.d.hp-ux
node.d.java
node.d.linux
node.d.netbsd
node.d.openbsd
node.d.sunos
```

is there a good reason to rename `sunos` to `solaris`?
Are there other opinions?

yunal:
I suspect `uname` still returns "sunos". Makefile.config:

```OSTYPE     := $(shell uname | LANG=C tr '[A-Z]' '[a-z]' | cut -f 1 -d _)```

shakemid:
In Makefile.config, `uname` is used and it returns 'SunOS'. But, I found `$^O` is used in Build.PL to determine OS name.  `$^O` returns 'solaris'.  I am not sure which is better.

> 237 File::Find::find( { wanted => \&_find_plugins_wanted },
> 238     'plugins/node.d', "plugins/node.d.$^O" );

It might be better to use `lc( qx/uname/ )` instead of `$^O` in Build.PL

niclan:
Yes, qx/uname/ in perl would be most portable then it seems.

shakemid:
I think so, too. I would like to close this pullreq.

PR #889, opened 2017-11-07
Update nfs plugins: Fixing autoconf to detect if service enabled

mobitux:
The autoconf function of the nfs plugins do not recognize the correspondent nfs service if it is not available.


sumpfralle:
Thank you!

PR #896, opened 2017-11-22
Fix argument size too big in threads plugin when there are too many processes.

robertabcd:
…rocesses.

### DIFF
--- a/plugins/node.d.linux/threads
+++ b/plugins/node.d.linux/threads
@@ -45,4 +45,6 @@ fi
 # be read.  It isn't entirely portable, but GNU grep should be a given on
 # Linux.  Sadly awk has no such equivalent option or we could skip grep
 # altogether
-grep -s '^Threads' /proc/[0-9]*/status | awk '{ sum += $2; } END { print "threads.value", sum; }'
+{ for fn in /proc/[0-9]*/status; do echo $fn; done; } | \
+    xargs -n 1000 grep -s '^Threads' | \
+    awk '{ sum += $2; } END { print "threads.value", sum; }'


plugins/node.d.linux/threads
```diff
@@ -45,4 +45,6 @@ fi
 # be read.  It isn't entirely portable, but GNU grep should be a given on
 # Linux.  Sadly awk has no such equivalent option or we could skip grep
 # altogether
-grep -s '^Threads' /proc/[0-9]*/status | awk '{ sum += $2; } END { print "threads.value", sum; }'
+{ for fn in /proc/[0-9]*/status; do echo $fn; done; } | \
```
sumpfralle: Technically you could omit the braces (`{...; }`) here, I think. Instead you could simply pipeline `done` and `xargs`.

For improved whitespace and quoting safety I would recommend the following instead:
```shell
find /proc/ -mindepth 2 -maxdepth 2 -type f -name status -print0 | \
    xargs -0 -r -n 1000 grep -s '^Threads:' | \
```

What do you think?

robertabcd: `find` did come to my mind. Is it guaranteed to have `find` on the system?

sumpfralle: > Is it guaranteed to have `find` on the system?

I am not aware of a strict definition of common requirements that munin plugins may rely on, but `find` would probably be a safe candidate for such a list.

sumpfralle:
Ping?

sumpfralle:
I added another commit including the "find" call on top of your commit. Thank you!

PR #897, opened 2017-11-27
Fast & ugly fix for PG10 support.

uu:
We're not interested in minor versions.


sumpfralle:
Are you aware of a problem, that could arise due to the missing check for the `PostgreSQL` string returned by `SELECT version()`?

Besides this I can see no objections for this change.

sumpfralle:
Since no objections were raised and another contribution from a different author (#919) relies on the above changeset, I assume it is an acceptable solution.

PR #906, opened 2018-01-31
Update plugin.rst : change refs from http to https for oss.oetiker.ch (rrdtool) web site

igit:
oss.oetiker.ch web site is now **https only** and does not redirect http to https


igit:
I'm gone to merge all changes in one patch

PR #909, opened 2018-02-14
Config.pm: fix anchor in URL in error message

jsavikko:
Fix an outdated anchor in URL to Wikipedia article in configuration error message "Hostname contains illegal characters".
This change to anchor was introduced by revision 744396074 (https://en.wikipedia.org/w/index.php?title=Hostname&diff=prev&oldid=744396074).


sumpfralle:
thank you!

PR #910, opened 2018-02-18
getversion: portability with bourne shell (e.g. Solaris)

sumpfralle:
The changes include:
* replace `$(...)` with backticks
* unify whitespace
* fix quoting issues
* fix or override (for portability) shellcheck warnings


sumpfralle:
btw: `local` is surely not supported by `/bin/sh` on Solaris, or?
(I would add this to the portability note in the beginning of the script)

PR #915, opened 2018-02-27
Handle negative values in spooled data

bcoppens:
When the value of spool data is negative the old regex would not match that line, which resulted in the code not timestamping negative values


sumpfralle:
Wow - that bug was hiding there for quite some time!

PR #919, opened 2018-03-02
Add PostgreSQL 10 support to postgres_xlog

crispygoth:
In PostgreSQL 10 the name of the WAL directory was changed so
pg_ls_dir('pg_xlog') no longer works.

Additionally, a new pg_ls_waldir function was added for specifically
listing the WAL directory. This function can be run by any user with the
pg_monitor role, so superuser access is no longer a requirement for this
check.

I've based my PR from master, although this does also require the fix from #897 in order to work correctly.

### DIFF
--- a/plugins/node.d/postgres_xlog
+++ b/plugins/node.d/postgres_xlog
@@ -36,6 +36,16 @@ Example :
   [postgres_pg92_*]
   env.PGPORT 5432
 
+Prior to PostgreSQL version 10, the user this plugin connects to the database
+as requires superuser privileges on the database:
+
+  ALTER ROLE your_nagios_user SUPERUSER;
+
+Since version 10 the user only requires the pg_monitor role:
+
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for keeping the plugin fresh!

> This function can be run by any user with the pg_monitor role, so superuser access is no longer a requirement for this check.

Should this changed requirement be documented (or just mentioned) somewhere? Or was this requirement just missing in the current state of the plugin?

Besides this, it looks good to me.

crispygoth:
It looks like the superuser requirement just wasn't documented previously. I've added a bit to the POD at the top of the plugin about the permission requirements for the different versions of PostgreSQL.

sumpfralle:
> I've added a bit to the POD at the top of the plugin about the permission requirements for the different versions of PostgreSQL.

Great - thank you!

PR #920, opened 2018-03-05
dev_scripts: "Build" parameters in 'install' for proper lib install path

sumpfralle:
Previously the Munin module files were partly installed outside of the
sandbox (here: ~/perl5/).

Thus Munin modules were not found when running code in the sandbox, e.g.:

 $ dev_scripts/install node
 Can't locate Munin/Node/Configure/PluginList.pm in @INC (you may need
 to install the Munin::Node::Configure::PluginList module) (@INC
 contains.
 BEGIN failed--compilation aborted at
 /home/foo/perl5/bin/munin-node-configure line 26.

### DIFF
--- a/dev_scripts/install
+++ b/dev_scripts/install
@@ -26,8 +26,10 @@ clean_sandbox() {
 
 make_sandbox() {
 	banner BUILD
-	perl Build.PL --install_base "${BASEDIR}/sandbox" --debug
-	./Build install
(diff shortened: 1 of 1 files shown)


sumpfralle:
I am not sure, why `./Build install` changes the path in `_build/runtime_params`.
I expected, the `--install_base` parameter used during the first `Build.PL` execution to be used for all subsequent runs.

Maybe someone is aware of a better solution for the misguided perl library installation path?

PR #926, opened 2018-03-11
Master::HTML: prevent access to static resources outside of "staticdir"

sumpfralle:
Currently it is possible to trick `munin-httpd` into delivering html files outside of the `staticdir`:
```shell
demo:~# curl -s "http://localhost:4948/static/..%2f..%2f../usr/share/apache2/icons/README.html" | md5sum
23091c19eaf031d01d86c3da50f3cfb8  -
demo:~# md5sum /usr/share/apache2/icons/README.html
23091c19eaf031d01d86c3da50f3cfb8  /usr/share/apache2/icons/README.html
```

A proxying webserver prevents this, since it canonicalizes the path before handing it over to its backend.

### DIFF
--- a/lib/Munin/Master/HTML.pm
+++ b/lib/Munin/Master/HTML.pm
@@ -30,6 +30,17 @@ sub handle_request
 	if ($path =~ m/static\/(.+)$/) {
 		# Emit the static page
 		my $page = $1;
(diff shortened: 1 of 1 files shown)


steveschnepp:
Nice catch. 

And, the way to handle it is also quite efficient.

PR #928, opened 2018-03-11
doc: describe new field-level attribute "unknown_limit"

sumpfralle:
The description of the *unknown_limit* attribute  is based on the code in `lib/Munin/Master/LimitsOld.pm`.

### DIFF
--- a/doc/reference/plugin.rst
+++ b/doc/reference/plugin.rst
@@ -429,6 +429,17 @@ Field name attributes
 
 ============
 
+.. _fieldname.unknown_limit:
+
+:Attribute: **{fieldname}.unknown_limit**
+:Value: positive integer
+:Type: optional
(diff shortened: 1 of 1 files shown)


doc/reference/plugin.rst
```diff
@@ -429,6 +429,17 @@ Field name attributes
 
 ============
 
+.. _fieldname.unknown_limit:
+
+:Attribute: **{fieldname}.unknown_limit**
+:Value: positive integer
+:Type: optional
+:Description: Defines the number of *unknown* values to be received successively, before the state of the dataset changes from *ok* to *unknown*. Use a higher value, if you want to tolerate a certain number of non-computable values, before an alarm should be raised. This attribute is available since Munin 3.0.
```
steveschnepp: Maybe just add "it gets reset every time you have a "ok" or "ko". 

Sounds logical once you think about it, but let's make it obvious.

sumpfralle: Good - I will add that.
What do you mean with "ko"?

PR #929, opened 2018-03-11
doc: describe datastructures (for developers)

sumpfralle:
This description of munin's data structures is targeted at developers. Thus it contains references to some code internals. It should be helpful for interested developers to understand from where to load and where to store data.


steveschnepp:
Nice, reminds me that there are some unused leftover columns.

sumpfralle:
> reminds me that there are some unused leftover columns.

:)

btw: do we detect data format changes?
I guess, it is reasonably safe to just throw away the state database whenever the data format (table and column specification) changes, or? Should we store a "data format version" or something similar in the settings table for detecting this?

PR #934, opened 2018-03-16
Simplify spooled value regex

### LINKED ISSUE #933: value regex mismatch in SpoolReader can lead to NO_SPOOLFETCH_DATA error
Debian 2.0.25-1+deb8u3 - also affects munin master ee5b5f8730070f7c65d7c1816c4ac3404e418831

The value regex in `Munin::Node::SpoolReader` does not match scientific notation. So if any plugins return such values, `spoolfetch` will return them without an `$epoch:` prefix. This leads `Munin::Master::Node::parse_service_config` into assuming that `$when = $now`, which may in turn bump the recorded spoolfetch timestamp to a value above anything returned by `munin-async`. Depending on timing, the next spoolfetch may happen before `munin-asyncd` has accumulated any data with a timestamp greater or equal than the stored timestamp, which will lead to the master logging a `NO_SPOOLFETCH_DATA` error in that run. Any data accumulated with a timestamp less than that is effectively lost.

The problem is mitigated by fixing the regex in SpoolReader. However I feel the server should be more robust in handling this scenario. Namely it should not consider `$now` timestamps for deriving `$spoolfetch_last_timestamp`, at the very least. I gave it a try, adding another boolean field `when_is_now` in `parse_service_config`, so as to later exclude the corresponding `when` from deriving the spoolfetch timestamp. I also wanted to try using the max of other data source timestamps, rather than `$now`. But then the question arises whether the timestamp to be used in RRD update should also change in such a case, and I did not want to affect that process.

zmousm:
Fix #933 (the source of the issue at least)

The regex could be further simplified (`^(\w+)\.value\s+(?:N:)?(.+)$`) since validating the value seems irrelevant at this point.

### DIFF
--- a/lib/Munin/Node/SpoolReader.pm
+++ b/lib/Munin/Node/SpoolReader.pm
@@ -156,7 +156,7 @@ sub _cat_multigraph_file
                 next;
             }
 
-            if (m/^(\w+)\.value\s+(?:N:)?(-?[0-9.]+|U)$/) {
+            if (m/^(\w+)\.value\s+(?:N:)?(.+)$/) {
                 $_ = "$1.value $epoch:$2";
             }
 


zmousm:
I updated the patch to use the more simple regex without validation, as suggested in #934.

sumpfralle:
I am sorry for pushing you back and forward :(
Could you please rebase it on master, instead? (despite my initial suggestion)

I was just reminded by the other folks on IRC, that we use the sane backport policy of merging into master and cherry-picking into stable-2.0.
I am sorry for having wasted your time with this back-and-forth.

zmousm:
no problem

sumpfralle:
Thank you!

PR #935, opened 2018-03-17
plugin mysql_: use SI unit for query_processing_time values

sumpfralle:
Previously the total time spent in queries was visualized in microseconds,
which prevented the usual SI related value scaling from working.

The internal storage format is not changed. Just a cdef is added.

Add optional format string handling for cdef fields.


niclan:
This looks most excellent.

PR #938, opened 2018-03-21
Let the server use max timestamp rather than now for non-timestamped dirtyconfig values

zmousm:
* Use special value `-1` rather than `$now` for `when` field in dirtyconfig output
  * Declare as a named constant
* Pre-compute `$last_timestamp`, in a separate iteration
  * Fall back to current timestamp
  * But not for spoolfetch timestamp
* If special value found in `when` field, use `$last_timestamp` instead
* `Munin::Master::UpdateWorker::_update_rrd_file` now returns number of
  RRD updates and `$last_timestamp` is an argument

### DIFF
--- a/master/lib/Munin/Master/Node.pm
+++ b/master/lib/Munin/Master/Node.pm
@@ -17,6 +17,9 @@ use Log::Log4perl qw( :easy );
 use Time::HiRes qw( gettimeofday tv_interval );
 use IO::Socket::INET6;
 
+# Used as a timestamp value, this declares none was found
+use constant NO_TIMESTAMP => -1;
+
 my $config = Munin::Master::Config->instance()->{config};
 
 # Quick version, to enable "DEBUG ... if $debug" constructs
@@ -359,7 +362,7 @@ sub parse_service_config {
         } elsif ($line =~ m{\A ([^\.]+)\.value \s+ (.+?) \s* $}xms) {
 	    $correct++;
 	    # Special case for dirtyconfig
-            my ($ds_name, $value, $when) = ($1, $2, -1);
+            my ($ds_name, $value, $when) = ($1, $2, NO_TIMESTAMP);
             
 	    $ds_name = $self->_sanitise_fieldname($ds_name);
 	    if ($value =~ /^(\d+):(.+)$/) {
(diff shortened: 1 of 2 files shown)


zmousm:
Rebased on master considering suggestion for #934

zmousm:
Kind bump...

Should I rebase on stable-2.0 ?

zmousm:
>   *  Which specific situation would this change improve? (Delayed fetching of data via munin-async? Or something else?)

It protects the master against data loss in spoolfetch, As noted in #940, master assumes `now` whenever it sees a non-timestamped value in dirtyconfig output, which also bumps spoolfetch timestamp, leading to potential data loss in subsequent runs (points with a lower timestamp). The patch decouples the latter from the former, only considering the max observed timestamp (i.e. never `now`) for updating the spoolfetch timestamp.

>   *  Regarding -1: could you use a named constant instead in order to make this special value easier to read and understand?

I also had reservations about giving a special meaning to -1, although it fit the use case. I will look into using a named constant.

>   *  Regarding 571:sub _update_rrd_files: this function is not used in 2.999 anymore (the code is there, but there is no caller). But this change would still be helpful for 2.0, or?

Yes it does apply to 2.0 as well. Let me get back to you about the different code path in 2.999.

> Regarding rebasing to 2.0 - this would indeed be great!

OK

zmousm:
> I am not familiar with this code, but at first sight it looks like the approach taken in this patch may no longer be plausible, due to stream-like processing (vs. all-at-once at the end) of spoolfetch data.

So my understanding is indeed that in 2.999:

* you can not pre-compute max timestamp and use that rather than `$now`,
* but at least you can avoid bumping `$last_timestamp` (and spoolfetch timestamp eventually) in such a case.

Do you have any comment? If not, I can open a different PR against master for this.

sumpfralle:
> Do you have any comment? If not, I can open a different PR against master for this.

Your reasoning sounds good and suitable.
Thus please go ahead and I will happily merge both changesets.
Thank you!

steveschnepp:
> stream-like processing

Yes, this was done for highly efficient ingestion, and lays the path for node-pushed updates in the future.

sumpfralle:
Sorry - somehow I forgot about this valuable proposal of yours (while its cousin #982 for 2.999 was merged back then).
Again: thank you!

PR #943, opened 2018-03-24
Plugin quota_usage: allow symlinks starting with real plugin name

sumpfralle:
Previously only symlinks starting with "quota-usage_" were supported.
Now symlinks may also start with the real plugin name ("quota_usage_").
This should avoid confusion.

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=730030


sumpfralle:
Any comments, please?
(my perl is rusty)

PR #944, opened 2018-03-27
Fixing things

steveschnepp:


lib/Munin/Master/Node.pm
```diff
@@ -720,16 +721,20 @@ sub _node_read_fast {
...
 
 sub _node_read {
-    my ($self) = @_;
+    my ($self, $trigger, $callback) = @_;
     my @array = ();
 
     my $line = $self->_node_read_single();
-    while($line ne ".") {
+    while($line = $self->_node_read_single()) {
+	last if $line eq ".";
```
sumpfralle: I am not familiar with the details, yet: is the dot a magic token?

lib/Munin/Master/UpdateWorker.pm
```diff
@@ -1147,13 +1147,18 @@ sub _create_rrd_file {
               "RRA:AVERAGE:0.5:288:450", # 450 days, resolution 1 day
               "RRA:MIN:0.5:288:450",
               "RRA:MAX:0.5:288:450");
-    }
-    elsif ($resolution eq 'huge') {
+    } elsif ($resolution eq 'huge') {
```
sumpfralle: I am relieved, that you do not seem to feel attached to the multi-line `} elsif` style :)

dev_scripts/install
```diff
@@ -35,7 +35,7 @@ EOF
 	if [ ! -z $XDG_RUNTIME_DIR ] && [ -r "$XDG_RUNTIME_DIR" ]
 	then
 		mkdir -p "$XDG_RUNTIME_DIR/munin-sandbox-var"
-		[ -e "$XDG_RUNTIME_DIR/munin-sandbox-var" ] || ln -sv "$XDG_RUNTIME_DIR/munin-sandbox-var" sandbox/var
+		[ -e "$XDG_RUNTIME_DIR/munin-sandbox-var" ] && ln -sv "$XDG_RUNTIME_DIR/munin-sandbox-var" sandbox/var
```
sumpfralle: shell nitpicking: if the last command in a block (e.g. `if ... fi`) is a compound command (especially `&&`), then the success of the block depends on this compound command. Thus when using exit-on-error, the `if` block would cause the script to exit for no obvious reason. (the compound command itself would not cause the exit-on-error - but the `if` block would do this)
Anyway: it is not important here - I just wanted to share a subtile problem of this structure.

PR #950, opened 2018-04-05
Specify minimum value for tuples

AMDmi3:
As far as I remember, rrdtool graphs need minimal value set for DERIVE counter to work correctly if the counts are reset to not produce a negative peak. However that's exactly what happens with postgres_tuples_ plugin:

image

I'm not 100% sure, but that could be because minimal values are not set by default. So set them.


sumpfralle:
Indeed - I think, you caught them all. Thank you!

AMDmi3:
Could this be merged to 2.0 as well?

sumpfralle:
Yes, it is good for 2.0, too!

Added in 567cd00ecdb88ebad5a65e9ba9fc91bfab6c7ac1.

AMDmi3:
Thank you!

PR #960, opened 2018-05-18
[postgres plugin] Do not set database=undef when paramdatabase.

bokutin:
From 7d58bca sudo -u munin munin-run postgres_locks_ALL now gives an error.

    host1 postgres[72156]: err [5-1] 2018-05-18 12:10:27 JST  pgsql  [local]  pgsql   3D000FATAL: database "pgsql" does not exist
    (pgsql is the default for user of freebsd ports.)

I am verifying that this will be as follows.
* postgres_locks_ALL is database=template1
* postgres_locks_mydb1 is database=mydb1


sumpfralle:
Thank you for cleaning up after me :)

PR #782, opened 2017-01-16
web: remove runtime dep to fonts.googleapis.com

steveschnepp:
Depending on some external ressource is bad as it makes us not self
contained anymore.

Either provide the resource locally, or find an alternative

PR #965, opened 2018-06-11
Remove external Fonts Dependency

sumpfralle:
* remove external font reference
* change font priorities to the ones used by bootstrap

Based on the discussion in #782.


chteuchteu:
Seems great! One font down, one remaining!

sumpfralle:
> 

> One font down, one remaining!

@chteuchteu: really? Let's get rid of the other one, too. Where is it hiding?

PR #967, opened 2018-06-12
Force the graph limits in snmp__if_multi when the speed is known

ls-initiatives:
This makes it much easier to spot busy/idle interfaces, particularly on devices with many interfaces such as stacked switches.

### DIFF
--- a/plugins/node.d/snmp__if_multi
+++ b/plugins/node.d/snmp__if_multi
@@ -648,7 +648,8 @@ sub do_config_if {
 
     print "graph_title $alias traffic\n";
     print "graph_order recv send\n";
-    print "graph_args --base 1000\n";
+    if ($speed) { print "graph_args --base 1000 --rigid --lower-limit -$speed --upper-limit $speed\n"; }
+    else { print "graph_args --base 1000\n"; }
     print "graph_vlabel bits in (-) / out (+) per \${graph_period}\n";
     print "graph_category network\n";
     print "graph_info This graph shows traffic for the \"$alias\" network interface.$extrainfo\n";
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for taking a look at the plugin!

I am not really sure, that this change would be beneficial for all users.
It would surely be good for ports with a good amount of traffic. But it would blur the information detail for lightly loaded links (and thus could hide relevant details), since all graphs would be zoomed up to the maximum link speed.

Maybe you could implement this change as a configurable option?
Or what do you think?

ls-initiatives:
I see your point.
Here are a few examples from my own switch.
This change makes it plain which ports were "active" that day, and which were not.
It also makes it plain that none of them are saturated.
On the other hand, fluctuations in the small activity get squashed indeed, so I agree that a 2x or maybe 10x increase would not be noticeable.
Not a problem for me but I see how this would be a pain for someone else.

I was thinking of making it logarithmic, but the we'd need to make the "receive" part positive (log does not accept negatives).
I think it would be good, but probably we'd want to make this throughout all the plugins.

if_1163-day
if_722-day
if_744-day

sumpfralle:
> I was thinking of making it logarithmic, but the we'd need to make the "receive" part positive (log does not accept negatives).

I think, a lot of people grew accustomed to the +/- visualization of network traffic. At least I did :)
I can hardly imagine, that we want to change this.

> I think it would be good, but probably we'd want to make this throughout all the plugins.

This makes this change even less probable, I am afraid.

What do you think: how should we proceed?
Do you want to make the enforcement of the upper limit visualization configurable for the `snmp__if` plugin?
Or maybe just rely on the warning thresholds for reporting intense port usage?

ls-initiatives:
Warnings have been very unreliable on my setup (way too many false alarms), maybe that's my root problem.

Adding an option would be fine by me, but I'll need time to add it.
Would you like to suggest a name for that option, for consistency ?

sumpfralle:
> Adding an option would be fine by me, but I'll need time to add it.

I would appreciate that!


> Would you like to suggest a name for that option, for consistency?

How about `graph_fixed_limit`? (defaulting to a false value)

Issue #914, opened 2018-02-25
Some issues trying out master


cgzones:
I am running a couple of issues when trying out the master branch:

## Alien::RRDtool is not installed
When setting munin up via `./dev_scripts/install 1` I am getting a prequesite miss:
```
Checking prerequisites...
  requires:
    !  Alien::RRDtool is not installed
```
  Is there a Debian package for this (I could not found any) and is this dependency needed?

## Node warning
When using `./dev_scripts/query_munin_node list` I am getting the following warning on the node: `Use of uninitialized value $node in exists at /path/to/sandbox/lib/perl5/Munin/Node/Server.pm line 337, <STDIN> line 1.`

## munin-cron fail
```
./dev_scripts/run munin-cron
Found: /path/to/sandbox/bin/munin-cron
Command line:  munin-cron
Global symbol "$onfield" requires explicit package name (did you forget to declare "my $onfield"?) at /path/to/sandbox/lib/perl5/Munin/Master/LimitsOld.pm line 442.
Global symbol "$onfield" requires explicit package name (did you forget to declare "my $onfield"?) at /path/to/sandbox/lib/perl5/Munin/Master/LimitsOld.pm line 442.
Global symbol "$onfield" requires explicit package name (did you forget to declare "my $onfield"?) at /path/to/sandbox/lib/perl5/Munin/Master/LimitsOld.pm line 443.
Global symbol "$onfield" requires explicit package name (did you forget to declare "my $onfield"?) at /path/to/sandbox/lib/perl5/Munin/Master/LimitsOld.pm line 444.
Global symbol "$onfield" requires explicit package name (did you forget to declare "my $onfield"?) at /path/to/sandbox/lib/perl5/Munin/Master/LimitsOld.pm line 445.
Global symbol "$onfield" requires explicit package name (did you forget to declare "my $onfield"?) at /path/to/sandbox/lib/perl5/Munin/Master/LimitsOld.pm line 448.
Global symbol "$onfield" requires explicit package name (did you forget to declare "my $onfield"?) at /path/to/sandbox/lib/perl5/Munin/Master/LimitsOld.pm line 449.
Global symbol "$onfield" requires explicit package name (did you forget to declare "my $onfield"?) at /path/to/sandbox/lib/perl5/Munin/Master/LimitsOld.pm line 452.
[...]

PR #975, opened 2018-06-29
Limits merge cleanup

sumpfralle:
The recent merge of stable into master seems to have caused at least a small problem manifesting as an undefined variable `$state` (which was removed meanwhile).

In the vicinity of this piece of code (in `LimitsOld.pm`) there were some more problematic parts, that are probably not fully migrated from state file to SQL database (to be expected - as discussed recently via IRC):
* some access to an `$onfield` hash (removed quite some time ago) - as mentioned in #914 (probably fixed now)
* the final state stored in the database (`my $new_state = $old_state;` - close to the end of `process_service`) looks wrong (`$old_state` does not change during this function and it was read from the database in the beginning)
    * should it be `$state` instead of `$old_state` (requiring some more changes)?
* there are some occurrences of `munin_set_var_loc` which are probably outdated?

In general: the code is a bit better now: it does not *always* spill the logs (and cron job mails) with warnings and probably it passes the autopkgtests of the Debian experimental packaging now. Thus I assume is it good enough to be merged.
But it probably needs a bit more work afterwards.

@steveschnepp: do you think, the limits code is worth to be incrementally improved now (by me)? Or would you rather want to rework this part of the code completely?

### DIFF
--- a/lib/Munin/Master/LimitsOld.pm
+++ b/lib/Munin/Master/LimitsOld.pm
@@ -364,6 +364,7 @@ sub process_service {
             $old_state, $old_num_unknowns) = $sth_state->fetchrow_array;
 
         $old_state ||= 'ok';
+        $old_num_unknowns ||= 0;
 
 		my $heartbeat = 600; # XXX - $heartbeat is a fixed 10 min (2 runs of 5 min).
 		if (! defined $current_updated_value || $current_updated_value eq "U") {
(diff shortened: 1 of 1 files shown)


sumpfralle:
@steveschnepp: do you have any opinion on these changes? I noticed, that they are necessary for running `munin-limits` with 2.999.9, thus I added these patches to the packaging, right now.

steveschnepp:
I didn't look closely at limits yet. I just know it needs some rework due to the lack of storable, and the presence of SQL

sumpfralle:
OK - I cherry-picked the commits now (for an easier merge).

PR #976, opened 2018-06-29
Makefile: Adding a warning in "make install"

steveschnepp:
The TGZ install is not the recommended way to install munin, as uninstalling
isn't implemented yet.

So this commit will warn users, and advise them to use a packaged version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,16 @@ tags:
 
 ######################################################################
 
-install: install-master-prime install-common-prime install-node-prime install-plugins-prime $(JAVA_INSTALL) install-man install-async-prime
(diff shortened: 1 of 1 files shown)


sumpfralle:
I *think* people downloading sources, preparing dependencies and running `make install` are used to programs offering only `install` and no `uninstall`. At least I would be surprised to see an `uninstall` target. Or did this become a usual feature during the last years?

Specifically for the Debian package this change (`make install-really` instead of `make install`) would complicate the process a bit, since the debhelper `dh_auto_install` calls this target by default.

sumpfralle:
Should we close it or do you insist? :)

steveschnepp:
Good idea only at first sight...

PR #977, opened 2018-06-30
Style checks

sumpfralle:
* fix many shellcheck issues of the current shell-based plugins
* add a "lint" target to the Makefile for style checks
    * for now: sh, bash, python
    * perl is currently missing, since the plugins are not perlcritic-clean
* add a travis step for this lint target
    * this should help contributors (and reviewers) to detect problems in advance

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+# settings for the "flake8" (python code style checks)
+[flake8]
+max-line-length = 99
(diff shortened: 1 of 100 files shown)


sumpfralle:
@steveschnepp: these changes can be easily merged into master later.
Do you have any comments?

PR #978, opened 2018-07-02
Update default warning threshold

alejandroscf:
As `recv` is a positve value, setting the warning threshold negative makes it launch an alarm when the value is actually OK

### DIFF
--- a/plugins/node.d/snmp__fc_if_.in
+++ b/plugins/node.d/snmp__fc_if_.in
@@ -129,7 +129,7 @@ if ($ARGV[0] and $ARGV[0] eq "config")
 	print "recv.graph no\n";
 	print "recv.cdef recv,8,*\n";
 	print "recv.max 4000000000\n";
-	print "recv.warning ", (-$warn), "\n" if defined $warn;
+	print "recv.warning $warn\n" if defined $warn;
 	print "send.label bps\n";
 	print "send.type DERIVE\n";
 	print "send.min 0\n";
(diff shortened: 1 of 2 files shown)


sumpfralle:
Good find!

The same structure is also used in `snmp__fc_if_`. Could you check, if there is the same problem and fix this too?
And please base your commits on `stable-2.0` instead of `master`. We are merging in this direction.

Thank you!

sumpfralle:
What do you think?

alejandroscf:
I'm sorry, I've been a bit busy.

Looks like the _fc_ module also could be corrected the same way, as it
assigns $warn and then use it with and without minus sign, so it will
generate the same fake alarms, but I can't test it as I don't have the
hardware.

Should I make a new pull request based in the stable-2.0 branch?

sumpfralle:
> Looks like the _fc_ module also could be corrected the same way, [..]

In this case, I am confident, that you should apply your change there, as well.

> Should I make a new pull request based in the stable-2.0 branch?

This is up to your taste :)
Personally I would probably prepare a different set of changes and force-push these into the same branch. Since this pull request is at your private discretion, history mangling is perfectly acceptable.

Thank you!

sumpfralle:
Ping?

alejandroscf:
Pong ;)

Just fixed snmp_fc_ plugin and rebased all changes to stable-2.0 branch

sumpfralle:
Cool - that was quick and final! :)

PR #980, opened 2018-07-03
Fix munin-asyncd fork bomb

zmousm:
If a forked child can't connect to munin-node, it should die rather than
moving to the next plugin. The latter leads to the child also forking
itself, recursively spawning an uncontrolled number of processes, which
quickly exhausts system resources.

Also applicable for stable-2.0.


zmousm:
OK, sure. A few months ago it was the other way around :)

sumpfralle:
> A few months ago it was the other way around :)

Variety is happiness :)

Thank you for patiently following our changing moods ...

Issue #940, opened 2018-03-21
Munin Master should not assume $now for non-timestamped dirtyconfig values


zmousm:
The server may receive non-timestamped values in dirtyconfig output. This may happen for example due to a regex mismatch in SpoolReader, as explained in #933.

The current timestamp is used in all such cases. While it makes sense to assume `$now` if a timestamp is missing, this will also bump `$last_timestamp` and `$spoolfetch_last_timestamp`, which can lead to `NO_SPOOLFETCH_DATA` in subsequent spoolfetch runs.

The server should try to avoid running into such problems, in any case. A more conservative approach is not to assume `$now` as a timestamp in such cases, but rather use the max of all timestamps returned. If that still doesn't cut it, then fall back to `$now`.

This is what I implemented in #938.

matt-matt2:
Aha, good call.
I believe i've run into something like this in earlier releases a few years back and hacked a way around it at the time by forcing spoolfetch timestamps to run five minutes out ..

I found that with long-running plugins (I mean 4.5 minutes or so) munin-asyncd records them in the spool with the time stamp from when the plugin _started_ running, not when it finished. So when the spoolfetch ran requesting records "newer than 5 minutes ago", there were none.

You may ask why the heck I had plugins running so long but .. well .. I did .. :)

sumpfralle:
Your proposals for 2.0 and 2.999 were merged as #938 and #982. Thank you!

PR #982, opened 2018-07-06
Do not bump spoolfetch timestamp from non-timestamped values

zmousm:
Master assumes `$now` whenever it sees a non-timestamped value in spoolfetch data, which also bumps spoolfetch timestamp, leading to potential data loss in subsequent runs (see #933 for example).

This patch decouples the latter from the former: It will not consider `$now` for deriving `$last_timestamp`, but it will still use it for storing the particular value. It would be better to use the max observed timestamp for this, before falling back to `$now`, but stream-like processing of spoolfetch data does not afford this (as discussed in #938).

This is a re-work of #938 for 2.999, both of which address issue #940.


sumpfralle:
Sorry for the long delay :(
Thank you!

PR #984, opened 2018-07-10
Fix paths on nutups_

bunder2015:
upsc is not always installed to /bin


plugins/node.d/nutups_
```diff
@@ -6,17 +6,18 @@
...
 
 UPS=$(basename $0 | cut -d_ -f2)
 FUNCTION=$(basename $0 | cut -d_ -f3)
+UPSC=$(which upsc)
 
 if [ "$1" = "autoconf" ]; then
-	[ -x /bin/upsc ] && [ -r /etc/nut/ups.conf ] && echo yes && exit 0
-	echo "no (/bin/upsc or /etc/nut/ups.conf not found)"
+	[ -x "$UPSC" ] && [ -r /etc/nut/ups.conf ] && echo yes && exit 0
+	echo "upsc or /etc/nut/ups.conf not found"
```
sumpfralle: `autoconf` expects a specific output pattern:
* success: `yes`
* failure: `no (some random reason)`

plugins/node.d/nutups_
```diff
@@ -43,7 +44,7 @@ voltages() {
 			echo "${i}.min 0"
 		done
 	else
-		upsc $UPS | sed -n '/volt/{
+		$UPSC $UPS | sed -n '/volt/{
```
sumpfralle: Please quote `"$UPSC"` in order to prevent whitespace issues.

See `shellcheck`, in case you want to take a look at more style issues :)

plugins/node.d/nutups_
```diff
@@ -11,22 +11,22 @@
...
 fi
 
 if [ "$1" = "suggest" ]; then
 	grep '^\[[^]]*\]$' /etc/nut/ups.conf \
 		| tr -d '][' \
-		| while read ups; do
+		| while read -r ups; do
 			for i in voltages freq charge current; do
-				echo ${ups}_${i}
+				echo "${ups}"_${i}
```
sumpfralle: Just a style note: you could have also quoted the full string up to the end - this is usually a bit easier to read.
But this is just a matter of taste ...

bunder2015:
I think I just made a mess. Let me figure out what I did wrong and I'll open a new PR.

sumpfralle:
Maybe I should not have said `rebase`: this can be complicated by git being too clever (or by us being a bit inconsistent about the directions of previous merge operations).
Probably you should just take the diff and apply it to the respective file (with an `.in` extension) in the `stable-2.0` branch.
Thank you!

PR #985, opened 2018-07-11
Fix paths on nutups_

bunder2015:
upsc is not always installed to /bin


sumpfralle:
That looks good - thank you!

PR #986, opened 2018-07-11
Update amavis : check if amavis is installed

barajus:
- check if the binary amavisd(openbsd) OR amavisd-new (debian) is installed.
- fix the code return if autoconf is no

### DIFF
--- a/plugins/node.d/amavis
+++ b/plugins/node.d/amavis
@@ -1,4 +1,6 @@
 #!/bin/sh
+# -*- sh -*-
+# vim: ft=sh
 
 : <<=cut
 
@@ -18,11 +20,11 @@ The configuration environment variables are available
 
 =item amavislog
 
-Path to logfile (Default: "/var/log/mail/mail.info")
+Path to logfile (Default: "/var/log/maillog")
 
 =item logtail
 
-Path to logtail command (Default: "logtail")
+Path to logtail command (Default: "/usr/local/bin/logtail")
 
 =back
 
@@ -31,7 +33,7 @@ plugin usually runs at.  To enable log reading it needs to run as a
 group or user that has read access, as shown above.
 
 By default the logtail program is started without any explicit path,
-but if it is not found in the system $PATH then you will need to specify
+but if it is not found in the system $PATH then you'll need to specify
 the full path for the program.
 
 =head2 EXAMPLE CONFIGURATION
@@ -39,27 +41,25 @@ the full path for the program.
 The following shows a typical configuration:
 
   [amavis]
-     env.amavislog     /var/log/mail/mail.info
-     env.logtail       /usr/bin/logtail
+     env.amavislog     /var/log/maillog
+     env.logtail       /usr/local/bin/logtail
+     env.amavisd       /usr/local/sbin/amavisd
      group adm
-
-=head2 DEFAULT CONFIGURATION
-
-This is the default configuration:
-
-  [amavis]
-    env.mktemp_command /bin/mktemp
 
 =head1 INTERPRETATION
 
-The plugin shows "probable spam", "surely spam" and "virus".  If your
+The plugin shows "probable spam", "surley spam" and "virus".  If your
 "probable spam" raises you may need to tune your spam configuration to
-classify more spam as "surely spam" and so be able to eliminate it.
+classify more spam as "surley spam" and so be able to elliminate it.
 
 =head1 MAGIC MARKERS
 
   #%# family=auto
   #%# capabilities=autoconf
+
+=head1 VERSION
+
+  $Id$
 
 =head1 BUGS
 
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for taking a look at this!

Regarding the exitcode for `autoconf`: it is really supposed to be zero - even for a failure. See the [documentation](http://guide.munin-monitoring.org/en/latest/develop/plugins/plugin-concise.html#autoconf) for details.

Regarding the autoconf check: indeed the current `autoconf` detection of amavis is way too simplistic. Your approach is better.

But I assume, that your goal is configurability for distributions, or?
In this case a single variable should be sufficient, since the plugin does not need to distinguish between the two implementations.
With a single variable for the location, a package maintainer (e.g. for Debian or a *BSD) would simply add a plugin configuration file that preempts the default location.

While you are at it: maybe you want to polish up the the autoconf evaluation?
In this case you, could use specific tests and output specific failure messages (e.g. `no (log file $AMAVIS_LOG not found)`) instead of the current single ANDed conditions with a generic failure message.

plugins/node.d/amavis
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh
+# -*- sh -*-
+# vim: ft=sh
```
sumpfralle: I share your editor preference, but I do not think, that we need to add these markers to the plugin files. Instead it can be done via private configuration.

plugins/node.d/amavis
```diff
@@ -18,11 +20,11 @@ The configuration environment variables are available
 
 =item amavislog
 
-Path to logfile (Default: "/var/log/mail/mail.info")
+Path to logfile (Default: "/var/log/maillog")
```
sumpfralle: I think, the old as well as the new value are both not universally applicable (at least here on Debian it would be `/var/log/mail.log`). Thus please leave this value unchanged if you do not have a good reason for a change.

plugins/node.d/amavis
```diff
@@ -18,11 +20,11 @@ The configuration environment variables are available
...
 
 =item logtail
 
-Path to logtail command (Default: "logtail")
+Path to logtail command (Default: "/usr/local/bin/logtail")
```
sumpfralle: I do not think, you should change this to a value that is only suitable for systems without a packaged `logtail`. At least here on my system it would work before (just using the system-wide configured PATH), but would fail with the new default.

plugins/node.d/amavis
```diff
@@ -31,36 +33,34 @@ plugin usually runs at.  To enable log reading it needs to run as a
 group or user that has read access, as shown above.
 
 By default the logtail program is started without any explicit path,
-but if it is not found in the system $PATH then you will need to specify
+but if it is not found in the system $PATH then you'll need to specify
```
sumpfralle: Back then I was taught in school, that this abbreviation is mainly suitable in spoken communication and not for written text :)
But you could convince me, if you are a native speaker and/or very sure, about what you are doing :)

plugins/node.d/amavis
```diff
@@ -31,36 +33,34 @@ plugin usually runs at.  To enable log reading it needs to run as a
...
 =head2 EXAMPLE CONFIGURATION
 
 The following shows a typical configuration:
 
   [amavis]
-     env.amavislog     /var/log/mail/mail.info
-     env.logtail       /usr/bin/logtail
+     env.amavislog     /var/log/maillog
+     env.logtail       /usr/local/bin/logtail
+     env.amavisd       /usr/local/sbin/amavisd
```
sumpfralle: Same as above: please do not change a default, without a good reason (i.e. it being more or less universally applicable).

plugins/node.d/amavis
```diff
@@ -31,36 +33,34 @@ plugin usually runs at.  To enable log reading it needs to run as a
...
-
-This is the default configuration:
-
-  [amavis]
-    env.mktemp_command /bin/mktemp
-
 =head1 INTERPRETATION
 
-The plugin shows "probable spam", "surely spam" and "virus".  If your
+The plugin shows "probable spam", "surley spam" and "virus".  If your
```
sumpfralle: My dictionary says *no* :)

plugins/node.d/amavis
```diff
@@ -31,36 +33,34 @@ plugin usually runs at.  To enable log reading it needs to run as a
...
 "probable spam" raises you may need to tune your spam configuration to
-classify more spam as "surely spam" and so be able to eliminate it.
+classify more spam as "surley spam" and so be able to elliminate it.
```
sumpfralle: Here my dictionary also disagrees :)

plugins/node.d/amavis
```diff
@@ -31,36 +33,34 @@ plugin usually runs at.  To enable log reading it needs to run as a
...
 
 =head1 MAGIC MARKERS
 
   #%# family=auto
   #%# capabilities=autoconf
 
+=head1 VERSION
+
+  $Id$
+
```
sumpfralle: Starting from this point I am quite convinced, that you copied an old version of the plugin over the one in the repository and simply committed this. This is not a good strategy :(

plugins/node.d/amavis
```diff
@@ -77,69 +77,75 @@ GPLv2
 
 =cut
 
-AMAVIS_LOG=${amavislog:-/var/log/mail/mail.info}
-LOGTAIL=${logtail:-logtail}
+mktempfile () {
```
sumpfralle: Same here: old code.

plugins/node.d/amavis
```diff
@@ -77,69 +77,75 @@ GPLv2
...
 if [ "$1" = "autoconf" ]; then
-        if [ -f "${AMAVIS_LOG}" -a -n "${LOGTAIL}" -a -x "${LOGTAIL}" ] ; then
-		echo yes
-		exit 0
-	else
-		echo no
-		exit 0
-	fi
+    if [ ! -f "${AMAVIS_LOG}" ]; then
+        echo "no (\$AMAVIS_LOG was not found)"
```
sumpfralle: Are you sure, that you want to escape the `$` character? Probably you want to display the real filename instead of confusing the user with the internal variable names used by this plugin.

sumpfralle:
@barajus: what do you think?

barajus:
Hi @sumpfralle.
Looks like I was working on the wrong version (I naively used the version from the package manager of my system).
I agree with your diff, do you need me to test it ?

sumpfralle:
I guess, all relevant proposed changes are merge in 5a77f6b. Thank you!

PR #990, opened 2018-07-17
Tests spoolfetch

steveschnepp:


t/config/munin.conf
```diff
@@ -1,7 +1,7 @@
-dbdir   /tmp
-htmldir /tmp
-logdir  /tmp
-rundir  /tmp
+dbdir   /dev/shm
```
steveschnepp: I know that `/dev/shm` is quite Linux specific, but as this is my main development environment it still makes sense.

I'd also welcome some patches that would enable some extra SQL to be issued after a connection, such as: `PRAGMA synchronous = OFF` and `PRAGMA journal_mode = MEMORY`.

These are useful in a testing context, but are harmful in a productive context. Therefore it should be configurable.

PR #995, opened 2018-08-07
LimitsOld: load configuration from file

sumpfralle:
While packaging munin 2.999.9 I noticed, that `munin-limits` does not seem to load the configuration file:

```
Use of uninitialized value in concatenation (.) or string at /usr/share/perl5/Munin/Master/LimitsOld.pm line 124.
Use of uninitialized value in concatenation (.) or string at /usr/share/perl5/Munin/Master/LimitsOld.pm line 128.
Could not open /munin-limits.lock for read/write: Permission denied
 at /usr/share/perl5/Munin/Common/Logger.pm line 216.
```

The proposed changes just add `my $config = Munin::Master::Config->instance()->{config};` (as seen in `lib/Munin/Master/Update.pm`.

I am not sure about the following details:
* Do we need to `use Munin::Master::Config` before?
* I think, this change is not sufficient for actually *parsing* the config file. This would probably require something like `$config->parse_config_from_file($conffile);` at the end of `limits_startup`. But I am confused, that I did not see this code pattern in other places.

### DIFF
--- a/lib/Munin/Master/LimitsOld.pm
+++ b/lib/Munin/Master/LimitsOld.pm
@@ -63,7 +63,7 @@ my $force          = 0;
 my $verbose        = 0;
 my $force_run_as_root = 0;
 my %notes          = ();
-my $config;
+my $config = Munin::Master::Config->instance()->{config};
 my $modified     = 0;
 my %default_text = (
     "default" =>


steveschnepp:
Munin limits needs some love, and be ported to SQL metadata.

steveschnepp:
In my design, only munin-update is really fully parsing the config files.

The "readers" like httpd and limits will leverage a preparsed config directly from the SQL db. 

Therefore they only need to parse config relevant to SQL connectivity.

sumpfralle:
> The "readers" like httpd and limits will leverage a preparsed config directly from the SQL db.

Without thinking about this deeper, I have the feeling that this could lead to a state of confusion (on the side of the admin), when configuration settings are changed in the file, but not active in a program started afterwards. But I am just thinking aloud. Simply proceed, if you see no problem ...

I will keep the patch of this changeset in the Debian experimental packaging for the time being (in order to use the proper file locations).

PR #996, opened 2018-08-07
Silence autoconf errors

cgzones:
Adjust several plugins, so that munin-node-configure does not complain about any.

### DIFF
--- a/plugins/node.d/squid
+++ b/plugins/node.d/squid
@@ -80,14 +80,15 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
 
 use strict;
 use Munin::Plugin;
+use Munin::Plugin::Framework;
 use IO::Socket::INET;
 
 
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for taking the time to fix issues of the `autoconf` procedure!

Most of your changes look good to me.

But the removal of the `autoconf` flag from the postgresql-related plugins is probably not correct, since these plugins use `Munin::Plugin::Pgsql`. This module handles `autoconf` requests.
Thus there is no need to remove the `autoconf` attribute, or?

And could you please apply your changes to the `stable-2.0` branch instead?

cgzones:
reworked the pgsql driver instead

cgzones:
I opened #1001  for the stable branch; this master pr in addition fixes the squid plugin, which is not existent in the stable branch

cgzones:
Removed all stable commits

sumpfralle:
cool - thanks!

PR #998, opened 2018-08-07
fix warning on node list operation

cgzones:
Use of uninitialized value $node in exists at .../sandbox/lib/perl5/Munin/Node/Server.pm line 332, <STDIN> line 1.


sumpfralle:
Superseded by #1002

PR #999, opened 2018-08-08
Improve database connection handling

sumpfralle:
* always use the configured database via `Munin::Master::Update::get_dbh`
* use separate instances for "dbh" and "dbh_state" for proper disconnects
* add some "disconnect" statements

### DIFF
--- a/lib/Munin/Master/HTML.pm
+++ b/lib/Munin/Master/HTML.pm
@@ -583,6 +583,7 @@ RENDERING:
 		use JSON;
 		print encode_json( \%template_params );
 	}
+	$dbh->disconnect();
 }
 
 sub _get_params_groups {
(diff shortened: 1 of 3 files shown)


sumpfralle:
cgzones notet in #914, that the above changes (specifically the excplicit `disconnect`) cause the following warnings:
```
DBI::db=HASH(0x55693a540c18)->disconnect invalidates 1 active statement handle (either destroy statement handles or call finish on them before disconnecting) at .../sandbox/lib/perl5/Munin/Master/HTML.pm line 586.
```

Indeed there are some places, where SQL statements are prepared in advance, before knowing, which ones will be used. We should probably `->finish()` all of them, just before `disconnect`.
Or what would be the proper approach?

steveschnepp:
statements should only need to be ->finish() if they where open via an ->execute().

lib/Munin/Master/HTML.pm
```diff
@@ -583,6 +583,7 @@ RENDERING:
 		use JSON;
 		print encode_json( \%template_params );
 	}
+	$dbh->disconnect();
```
steveschnepp: This isn't needed, as it will be done automatically when `$dbh` gets out of scope.

And, if you want to be explicit, you forgot an occurence above (when replying 404). So better leave it to RAII.

sumpfralle: Indeed: I was not aware of the implicit resource cleanup.
Should I remove all the "disconnect" statements again in order to not pollute the exemplary character of the code?

steveschnepp: Yes, please 👍

PR #1002, opened 2018-08-08
fix warning on node list operation

cgzones:
stable pr of #998

### DIFF
--- a/node/lib/Munin/Node/Server.pm
+++ b/node/lib/Munin/Node/Server.pm
@@ -168,7 +168,7 @@ sub _process_command_line {
 
     logger ("DEBUG: Running command '$_'.") if $config->{DEBUG};
     if (/^list\s*([0-9a-zA-Z\.\-]+)?/i) {
-	my $hostname_lc = defined($1) ? lc($1) : undef;
+	my $hostname_lc = defined($1) ? lc($1) : '';
         _list_services($session, $hostname_lc);
     }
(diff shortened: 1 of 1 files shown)


sumpfralle:
looks good, thank you!

cgzones:
I do not know

There are some differences in `_list_services` between master and stable regarding the `$node` usage:

stable
```
my ($session, $node) = @_;
$node ||= $config->{fqdn};
if (exists $nodes{$node}) {
    my @services = @{$nodes{$node}};
```

master
```
my ($session, $node) = @_;
if (keys %nodes == 1 && ! exists $nodes{$node}) {
    # Only one node. Naming mismatch. Just give the use what he wants.
    ($node) = keys %nodes;
}

if (exists $nodes{$node}) {
    my @services = @{$nodes{$node}};
```

Maybe the line ```$node ||= $config->{fqdn};``` needs to be reintroduced?

sumpfralle:
> Maybe the line $node ||= $config->{fqdn}; needs to be reintroduced?

Yes, indeed I think, this is the proper behaviour for an `undef` input: we really want to get rid of the possibility of `undef` as early as possible.

Thus we could revert the above change for `stable-2.0` and reintroduce the above line to `master`.

Do you agree?

cgzones:
Yes, this would make sense

sumpfralle:
Done - also see a3ca1531.

cgzones:
Can you revert the `undef` to `''` change?

PR #1003, opened 2018-08-08
drop Alien::RRDtool dependency

cgzones:
this module is nowhere used and this fixes

  *************************************************
  *  BUILD
  *************************************************
  Checking prerequisites...
    requires:
      !  Alien::RRDtool is not installed

  ERRORS/WARNINGS FOUND IN PREREQUISITES.  You may wish to install the versions
  of the modules indicated above before proceeding with this installation

  Run 'Build installdeps' to install missing prerequisites.

### DIFF
--- a/Build.PL
+++ b/Build.PL
@@ -122,7 +122,6 @@ my $build = Module::Build->new(
         'Net::Server::Fork'      => '0',
         'Parallel::ForkManager'  => '0',
         'Params::Validate'       => '1.08',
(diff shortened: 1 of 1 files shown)


sumpfralle:
This is a bit of mystery, for me, too. I *think*, it would be a way to install rrdtool (indirectly via cpan). Maybe this hint could be useful for developers on distributions without an rrdtool package (if there is any).

@steveschnepp: do you have an idea, whether this dependency is useful?

PR #1004, opened 2018-08-08
add devscript for starting munin httpd daemon

cgzones:


sumpfralle:
I am not sure, that another script for this purpose would be justified, since it is just the same as `dev_scripts/run munin-httpd`. You could even leave out all the sourcing, `cd` and the `set` commands.

Maybe just add a README.md file to the directory, pointing to the [sandbox development section of the guide](http://guide.munin-monitoring.org/en/latest/develop/environment.html#run-munin-in-a-sandbox).

What do you think?

cgzones:
Yes, I agree

PR #1005, opened 2018-08-08
fix warning on munin-update

cgzones:
Use of uninitialized value $ident in concatenation (.) or string at /home/christian/Coding/workspace/munin/sandbox/lib/perl5/Munin/Master/Update.pm line 175.

### DIFF
--- a/lib/Munin/Master/Update.pm
+++ b/lib/Munin/Master/Update.pm
@@ -174,7 +174,7 @@ sub _run_workers {
 
 	WORKER_LOOP:
 	for my $worker (@{$self->{workers}}) {
-		my $worker_pid = $pm->start();
+		my $worker_pid = $pm->start($worker);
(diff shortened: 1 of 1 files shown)


sumpfralle:
Just curious: removing the `$ident` parameter from the `run_on_finish` hook would also fix the issue, or?

Which approach would you consider to be preferable?

Thank you!

cgzones:
Yes, you are right. I thought the ident would maybe be helping debugging but I have no real preference.

sumpfralle:
If you think, the `$worker` is a suitable debugging identification (I do not know the code around so much), then it would be ok to leave it. Otherwise I would prefer to see it removed.

I leave it up to you to decide. Thank you!

cgzones:
In that case I like passing `$worker` as ident more

PR #1008, opened 2018-08-10
[dev_scripts] remove unavailable plugins on ./install node

cgzones:


sumpfralle:
Good catch! Thank you!

PR #1009, opened 2018-08-10
overlook some plugins

cgzones:


plugins/node.d/df
```diff
@@ -24,7 +26,7 @@ The following environment variables are used by this plugin:
 . "$MUNIN_LIBDIR/plugins/plugin.sh"
 
 
-df_entries=$(df -P -l 2>/dev/null | sed -e 1d -e '/\/\//d' -e 's/%//' | sort)
+df_entries=$(df -P -l 2>/dev/null | sed -e '1d' -e '/\/\//d' -e 's/%//' | sort) || true
```
sumpfralle: Do you have a specific error in mind, that should be tolerated in this line?
I can only imagine a failed `df` invocation. And I guess, we would not want to ignore that?

plugins/node.d/df_inode
```diff
@@ -30,7 +32,7 @@ The following environment variables are used by this plugin:
 . "$MUNIN_LIBDIR/plugins/plugin.sh"
 
 
-df_entries=$(df -P -l -i -x btrfs 2>/dev/null | sed -e '1d' -e '/\/\//d' -e 's/%//' | sort)
+df_entries=$(df -P -l -i -x btrfs 2>/dev/null | sed -e '1d' -e '/\/\//d' -e 's/%//' | sort) || true
```
sumpfralle: See above.

sumpfralle:
What a nice cleanup!

Please take a look at my two nitpicks below.

Besides this: could you please apply these changes to the `stable-2.0` branch instead? Thank you!

plugins/node.d/df_inode
cgzones: we do not want to terminate the script here (with the introduction of `set -e`)

plugins/node.d/df
cgzones: without the quotes it does not work in my sandbox (btw. df_inode has the quotes too)

sumpfralle: The quotes are fine (but they should not make a difference). I meant `|| true`.

plugins/node.d/df_inode
sumpfralle: Yes, I understand. But I cannot imagine, which part of this line could fail (for a good reason that we want to ignore). What do you think?

cgzones:
You're right; updated

sumpfralle:
Good!
Could rebase (or apply) your changes on `stable-2.0`?

btw.: feel free to just force-push your new branch onto an old instead of opening a new pull request. But this is just a matter of taste ...

cgzones:
opening a new one seems to be easier for me (see #1016)

PR #1010, opened 2018-08-10
[plugins] rename conflicting linux plugins

cgzones:


cgzones:
You're right; the renaming is not necessary (while playing around I thought it might be problematic)

Updated

sumpfralle:
Great - thanks!

PR #1012, opened 2018-08-10
 [munin-node-configure] fail on non-zero autoconf exit

cgzones:
all plugins should match this behavior now


sumpfralle:
I agree, the maintained plugins should be ready for this change.
(the lines to be removed are now almost nine years old)

Please apply your change on `stable-2.0`.

sumpfralle:
Merged #1013 into `stable-2.0`

PR #1013, opened 2018-08-11
[munin-node-configure] expect autoconf to return 0 even in the no case

cgzones:
plugins should nowadays satisfy this


sumpfralle:
Thank you!

PR #1015, opened 2018-08-11
[plugins] fix autoconf for lpstat

cgzones:


sumpfralle:
There is no escape from your cleanup :)
Cool!

PR #1017, opened 2018-08-11
[munin-node-configure] do not error out on empty suggest output

cgzones:


script/munin-node-configure
```diff
@@ -372,7 +372,7 @@ sub run_plugin
 
     # Ignore debug output
 	my @response = grep !/^#/, @{ $res->{stdout} };
-    $plugin->log_error('Nothing printed to stdout') unless (scalar @response);
+    $plugin->log_error('Nothing printed to stdout') unless (scalar @response || $mode eq 'suggest');
```
sumpfralle: I think, you want to omit the error message in case of an empty response, if the `suggest` was requested. Maybe a comment would help to understand your fellow code readers here.
btw.: the spacing in the line above looks different - maybe just fix that, too?

lib/Munin/Node/Configure/Plugin.pm
```diff
@@ -254,10 +254,11 @@ sub parse_suggest_response
         }
     }
 
-    unless (@{ $self->{suggestions} }) {
-        $self->log_error("No valid suggestions");
-        return;
-    }
+    # ignore empty suggest response
```
sumpfralle: Please do not leave dead code in comments around :)

Is this change necessary combined with the change below?
Maybe add some flag, whether an empty response should be ignored? (just guessing - I did not take a closer look)

steveschnepp: > dead code in comments

That's indeed always bad in commited code.

cgzones:
see #1020

PR #1024, opened 2018-08-12
[plugins] fix procfs swap

cgzones:
/proc/stat does not contain any swap entires
when the vmstat value is zero it gets overridden

note: procfs seems not to be in stable-2.0

### DIFF
--- a/plugins/node.d.linux/procfs
+++ b/plugins/node.d.linux/procfs
@@ -246,7 +246,7 @@ if ( -r "/proc/swaps" ) {
 	 type  => "DERIVE",
 	 min   => 0,
 	 graph => "no",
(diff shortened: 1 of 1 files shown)


sumpfralle:
I am not sure, whether the `swap` field was removed somewhen from `/proc/stat`. At least http://man7.org/linux/man-pages/man5/proc.5.html still documents it. Or maybe it is platform-specific? I cannot tell.

How about using a selector (`defined($vmstat{pswpin}) ? ... : ...`) instead in order to only override the value if it is really missing?

cgzones:
Somehow odd..
updated

sumpfralle:
> Somehow odd..

Indeed.

Thanks!

PR #1029, opened 2018-08-13
[plugins] remove linux specific df plugin

cgzones:
The linux df plugin uses multigraph for displaying the different devices.
This means that for each device a new graph is generated.
The standard df plugin combines all devices like the df_inode plugin into one single graph.
I think this is a more compact way.

p.s.: currently the multigraph support is also broken, cause a dot instead of an underscore is used as seperator.


steveschnepp:
It would be better to convert it to a hierarchical plugin than to remove it

cgzones:
But what is then the advantage of this plugins in comparison to the generic df one?

cgzones:
Maybe rename the linux `df` plugin to something like `df_detailed` (so that the default `df` plugin is also installed) and set the family to `contrib` (so that it's not enabled by default)?

sumpfralle:
While I appreciate trivial and non-conditional plugins, I very much dislike code duplication. Thus I would really prefer to avoid adding another copy of existing code with slight deviations :(

sumpfralle:
Closing in order to avoid code duplication.

PR #1031, opened 2018-08-14
Some manual/contrib related plugin changes

cgzones:
see commits for details

### DIFF
--- a/lib/Munin/Node/Configure/Plugin.pm
+++ b/lib/Munin/Node/Configure/Plugin.pm
@@ -75,6 +75,15 @@ sub suggestion_string
         # Report why it's not being used
         $msg = " [$self->{defaultreason}]";
     }
+    elsif (! $self->{capabilities}->{autoconf} && ! $self->{capabilities}->{suggest}) {
+        $msg = " [[[ plugin has neither autoconf not suggest support ]]]";
+    }
+    elsif ( scalar @{$self->{errors}} != 0 ) {
+        $msg = " [[[ plugin has errors, see below ]]]";
+    }
+    else {
+        $msg = " [[[ plugin gave no reason why ]]]";
+    }
 
     return $self->{default} . $msg;
 }
@@ -134,7 +143,7 @@ sub _suggested_links
 {
     my ($self) = @_;
 
-    # no suggestions if the plugin shouldn't be installed 
+    # no suggestions if the plugin shouldn't be installed
     return [] if $self->{default} ne 'yes';
 
     if ($self->is_wildcard or $self->is_snmp) {
@@ -400,7 +409,7 @@ directory, 'no' otherwise.
 =item B<suggestion_string()>
 
 Returns a string detailing whether or not autoconf considers that the plugin
-should be installed.  The string may also report the reason why the plugin 
+should be installed.  The string may also report the reason why the plugin
 declined to be installed, or the list of suggestions it provided, if this
 information is available.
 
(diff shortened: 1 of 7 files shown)


lib/Munin/Node/Configure/Plugin.pm
```diff
@@ -75,6 +75,16 @@ sub suggestion_string
         # Report why it's not being used
         $msg = " [$self->{defaultreason}]";
     }
+    elsif (! $self->{capabilities}->{autoconf}) {
```
sumpfralle: I think, `autoconf` is not a strict requirement for `suggest`, or?
Thus I think, this message would be misleading.

lib/Munin/Node/Configure/Plugin.pm
```diff
@@ -75,6 +75,16 @@ sub suggestion_string
...
+        # No autoconf support
+        $msg = " [[[ plugin has no autoconf support ]]]";
+    }
+    elsif ( scalar @{$self->{errors}} != 0 ) {
+        $msg = " [[[ plugin has errors, see below ]]]";
```
sumpfralle: Do you have something specific in mind with the three square brackets? The previous message above uses just one - probably you can stick to this pattern here, too?

lib/Munin/Node/Configure/Plugin.pm
cgzones: I added `&& ! $self->{capabilities}->{suggest}`
The idea is that when no autoconf is supported the default `no` value does not get overridden by a plugin run

lib/Munin/Node/Configure/Plugin.pm
cgzones: I wanted to make it distinguishable if the plugins supplied a message or munin itself has something to say; but I can change it

lib/Munin/Node/Configure/Plugin.pm
sumpfralle: OK. Sorry - my first attempt of understanding that change failed :(
Now I see, that the function is called `suggestion_string`, but indeed it is used for `autoconf` (and not for `suggest`).

Thus you should ignore my previous comment.

Since I do not really grasp the flow of the `suggestion_string` function: could you please explain, which problem your change solves?
I am not a user of that plugin module and I have problems understanding what the fields `default` and `defaultreason` are supposed to do and why we should not check for `{capabilities}->{autoconf}` straight at the beginning of the function (or even before calling it).

cgzones: `suggestion_string` is called from `munin-node-configure` when using the `--suggest` flag.
`->default` is the autoconf answer ("yes"/"no") and `->defaultreason` the autoconf description. (e.g. `echo "no (command foo not found)"` results in `default` equals "no" and `defaultreason` equals "command foo not found")

E.g.:
```
$ ./dev_scripts/install node
$ ./dev_scripts/run munin-node-configure --suggest --families=manual
Found: ...munin/sandbox/bin/munin-node-configure
Command line:  munin-node-configure --suggest --families=manual
Plugin                     | Used | Suggestions                            
------                     | ---- | -----------                            
apt                        | no   | yes                                    
apt_all                    | no   | yes                                    
bind9_rndc                 | no   | no
...
```

With my change, the `no` is further explained:

```
./dev_scripts/run munin-node-configure --suggest --families=manual
# [dev/run] Found: .../munin/sandbox/bin/munin-node-configure
# [dev/run] Command line:  munin-node-configure --suggest --families=manual
Plugin                     | Used | Suggestions                            
------                     | ---- | -----------                            
apt                        | no   | yes                                    
apt_all                    | no   | yes                                    
bind9_rndc                 | no   | no [[[ plugin has neither autoconf not suggest support ]]]
...
```

cgzones:
see #1036

PR #1033, opened 2018-08-17
Merge jenkins plugins

sumpfralle:
Merge the separate jenkins plugins into a single multigraph plugin in order to reduce duplication.

The data files of the previous separate plugins are re-used, if the new plugin is symlinked as `jenkins`. Otherwise (with its new longer name - avoiding pollution of the namespace), the names of rrd files would differ.

The merged plugins offer additional configuration variables (path of the jenkins directory and a regex for filtering running build processes).


sumpfralle:
@h01ger: maybe you want to take a look, if you notice any problems or changes?

btw: The current graphs for build results [look like they could be improved](https://jenkins.debian.net/munin/debian.net/jenkins.debian.net/jenkins_builds_results.html) by switching to a nested multigraph:
* the top level would be the same as the current [results summary graph](https://jenkins.debian.net/munin/debian.net/jenkins.debian.net/jenkins_builds_results_summary.html)
* the lower level would consist of one graph per group - each containing just the different result states for this group.
I think, this would improve the visualization. Should I do that? What do you think?

h01ger:
looks good to me, thanks!

sumpfralle:
OK - so I transformed the "build results" into a pretty multigraph plugin.
See 6afb8ca301d5c84457fd504133a0bf8710a09fac for a note regarding history keeping, if necessary.

PR #1035, opened 2018-08-17
munin-node-configure: prepare for taint mode and fix help message

cgzones:


script/munin-node-configure
```diff
@@ -328,7 +328,14 @@ sub init_snmp
 sub run_plugin
 {
     my ($plugins, $plugin, $mode) = @_;
-    my $name = $plugin->{name};
+
+    my $name;
+    if ($plugin->{name} =~ /^([\w]+)$/x) {
```
sumpfralle: I looked through the plugins in `stable-2.0`. There seem to be at least two plugins containing dashes. Thus maybe you should add this character to the regex?
btw.: what is the purpose of this conditional? Maybe add a comment above?

cgzones:
I do not know if its a good idea to merge it into stable, cause playing with the taint mode might break things (I hope however it does not)

cgzones:
The help message is fixed, cause `Pod::Usage::pod2usage` looks for the pod documentation `OPTIONS` not `GENERAL OPTIONS`.

dev_scripts/install
```diff
@@ -33,6 +33,9 @@ make_sandbox() {
 	# to _build/runtime_params and be used for munin module installation.
 	./Build --install_base "${BASEDIR}/sandbox" install
 
+	# disable taint mode in sandbox
```
sumpfralle: Could you explain (and maybe comment in the code), why the taint mode should be disabled in the sandbox environment? Wouldn't this be a good place to be informed about tainting issues early in development?

script/munin-httpd
```diff
@@ -24,9 +24,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 =cut
 
-# Trust PERL5LIB from environment
```
sumpfralle: I cannot really comment on why these lines where here, but could you explain, why they should go away?

script/munin-httpd
```diff
@@ -61,7 +58,5 @@ sub handle_request
 
 package main;
 
-$ENV{PATH} = '/usr/bin:/bin';
```
sumpfralle: Same here: I am not sure, why the PATH was limited up to now. Probably preventing unwanted locations of programs to be executed directly?
I have the feeling, that the current restriction is not useful (i.e. would be a surprise for installations with non-packaged installations of rrdtool or similar). But I am not sure about its purpose.

script/munin-node-configure
```diff
@@ -30,6 +30,7 @@ my $config = Munin::Node::Config->instance();
 my @all_families     = qw/auto manual contrib/;
 my @default_families = qw/auto/;
 
+$ENV{PATH} = '/sbin:/bin:/usr/sbin:/usr/bin';
```
sumpfralle: This reduced (or extended?) PATH would be used by all plugins run via `munin-node-configure`, or?
This could be a surprise, if the local admin wants to make non-packaged installed programs accessible to plugins?

sumpfralle:
Thank you for your detailed explanation!

I am not too familiar with perl's taint mode, thus I had a look at `man perlsec`.

As far as I understand, there are two issues to be solved when enabling taint mode:
1) make sure, that every component in PATH is an absolute path and is only writable by its owner
2) PERL5LIB is ignored completely

Issue (1) should be no problem, I think, since we use an absolute path for the sandbox and it is only writable by its owner, I guess?

Issue (2) prevents us from using the taint mode combined with the development sandbox, since `PERLLIB` is ignored completely in taint mode. Thus we need to disable taint mode in all our scripts in the sandbox. Correct?

Thus your changes do the following:
* enable taint mode for all munin executables when installed
* disable taint mode for all munin executables in sandbox mode (since we need PERLLIB)
* remove "untaint" code in `munin-httpd` that was probably introduced in order to get it working in the sandbox

If my understanding is correct, then I welcome the above changes, but please add a bit of comment to the "disable taint mode in sandbox" change in order to clarify for future readers, that this is necessary due to our use of PERLLIB.

One last thing: you add a PATH definition in `munin-node-configure`. Is this really necessary?

sumpfralle:
> You gave me an idea: I kept the taint mode active in the sandbox and modified therefore the run script so it calls all binaries with an -Mlib parameter for the needed sandbox libs.

cool!

> the new path in m-n-c is needed cause m-n-c calls net::domain::hostfqdn which calls the binary domainname

But our PATH is *sane* (i.e. only absolute paths, no non-owner-writable directories), so it should not be necessary? (Just guessing - I did not try it)

cgzones:
Somehow it does not work :(

sumpfralle:
I just cloned your branch `master_mnc` and commented out the PATH overrides:

```shell
$ grep -r PATH sandbox/bin/
sandbox/bin/munin-run:#$ENV{PATH} = '/usr/bin:/bin';
sandbox/bin/munin-node:#$ENV{PATH} = '/usr/bin:/bin';
sandbox/bin/munin-node-configure:#$ENV{PATH} = '/usr/bin:/bin';
```

Afterwards I was able to run the following without error messages:
* `dev_scripts/start_munin-node`
* `dev_scripts/run munin-run df`
* `dev_scripts/run munin-node-configure`

Thus maybe you have a customized PATH variable when running the sandbox schripts?
A proper solution for this could to specify the `PATH` variable in `dev_scripts/common.sh` - or what do you think?

steveschnepp:
> I do not know if its a good idea to merge it into stable, cause playing with the taint mode might break things (I hope however it does not)

Merging in stable is a no-go for me, as we should try hard to make only fixes on productive code. Improvements are possible when the current code is outdated enough to be useless. 

Example: a plugin that supports an old version that isn't shipped anymore or upgrading to a currently-supported version of Perl.

script/munin-httpd
steveschnepp: `PERL5LIB` is a trick akin to `LD_LIBRARY_PATH`, but for Perl. 

This should not be used in a daemon listening on a TCP port as UID=0.

script/munin-httpd
steveschnepp: Exactly.

sumpfralle: Do you mean, we should remove the PATH override?
(I would think so)

sumpfralle:
>> Regarding the improvement of the pod headers: great! I think, they should go to stable-2.0, too.
>
> I'd vote for "no", sorry

Specifically it is about #1039. Do you consider this change to be problematic?

script/munin-httpd
steveschnepp: Yes.

cgzones:
I cleaned the history a bit up, removed the explicit $ENV{PATH} additions and drop the deletion in munin-node mentioned by steveschnepp

cgzones:
anything new here; I think the requested change is applied

sumpfralle:
@steveschnepp: do you have any more comments?
I think, everything is fine.

sumpfralle:
Thank you, cgzones!

PR #1039, opened 2018-08-18
[mnc] fix documentation so that --help works

cgzones:


sumpfralle:
btw: #1035 contained more documentation-related changes. Shouldn't they go to `stable-2.0`, too?

cgzones:
i do not think so, the AUTHOR -> AUTHORS and LICENSE -> COPYRIGHT changes apply only at master

sumpfralle:
Good - thanks for clarifying!

PR #1047, opened 2018-08-19
add perlcritic makefile target and fix/silence some issues

cgzones:
currently perlcritic is only called with severity 5 and not for the folders `script` and `lib/Munin/Master`, cause this is causing to much errors by now

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,11 @@ install: $(BUILD_SCRIPT)
 		-e "$$(perl -I lib -M"Munin::Common::Defaults" \
 			-e "Munin::Common::Defaults->print_as_sed_substitutions();")"
 
+.PHONY: perlcritic
+perlcritic:
+	#TODO: apply to scripts/ and lib/Munin/Master/
+	perlcritic lib/Munin/Node
+
 .PHONY: lint
 lint:
 	@# SC1008: ignore our weird shebang (substituted later)
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,7 @@ install: $(BUILD_SCRIPT)
 		-e "$$(perl -I lib -M"Munin::Common::Defaults" \
 			-e "Munin::Common::Defaults->print_as_sed_substitutions();")"
 
+#TODO: merge with lint target, when done
 .PHONY: perlcritic
 perlcritic:
 	#TODO: apply to scripts/ and lib/Munin/Master/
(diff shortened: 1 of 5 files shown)


lib/Munin/Node/Utils.pm
```diff
@@ -20,7 +20,8 @@ sub set_difference
     my %set;
     @set{@$A} = ();
     delete @set{@$B};
-    return sort keys %set;
+    my @tmp = sort keys %set;
```
sumpfralle: I am not too deep into perl, thus please excuse my ignorance: wouldn't `return sort keys %set` work, as well?

lib/Munin/Node/Server.pm
```diff
@@ -216,6 +216,7 @@ sub _process_command_line {
 sub _get_commandline {
   my $self = shift;
 
+  ## no critic Variables::ProhibitPunctuationVars
```
sumpfralle: How about following the recommendation of `man Perl::Critic::Policy::Variables::ProhibitPunctuationVars` and using `English` and `$PROGRAM_NAME`?
(I am not aware of side-effects of `English`)

Makefile
```diff
@@ -54,6 +54,11 @@ install: $(BUILD_SCRIPT)
 		-e "$$(perl -I lib -M"Munin::Common::Defaults" \
 			-e "Munin::Common::Defaults->print_as_sed_substitutions();")"
 
+.PHONY: perlcritic
```
sumpfralle: I would prefer to keep this in the `lint` target and maybe mask its error code until all issues are fixed?
Or maybe at least add a TODO for merging these targets, when the issues are fixed?

lib/Munin/Node/Utils.pm
cgzones: Actually yes, but https://metacpan.org/pod/release/THALJEF/Perl-Critic-1.122/lib/Perl/Critic/Policy/Subroutines/ProhibitReturnSort.pm

sumpfralle: Great - thank you for clarifying!

lib/Munin/Node/Server.pm
cgzones: done

Makefile
cgzones: added

sumpfralle:
Great - thanks!

PR #1049, opened 2018-08-20
fix more perlcritic warnings

cgzones:
now all code complies to severity 5


sumpfralle:
Perfect!
`English` makes such a difference ...

I assume, the `perlcritic` run is free of errors now? So how about moving it to the `lint` target now?

sumpfralle:
You also fixed some shellcheck issues in the plugins now.
As far as I am aware, some of these are partly fixed in `stable-2.0` and the replacement of the jenkins plugins is pending in #1033.
Thus please let us keep this changeset simple and leave out the plugins for now.
Thank you!

lib/Munin/Master/Graph.pm
```diff
@@ -667,26 +664,26 @@ sub RRDs_graph {
...
+	#my $chld_pid = open3($chld_out, $chld_in, $chld_err, "rrdtool", "graphv", @_);
+	#
+	#DEBUG "[DEBUG] RRDs_graph(chld_out=".${$chld_out->string_ref}.")";
+	#DEBUG "[DEBUG] RRDs_graph(chld_err=".${$chld_err->string_ref}.")";
+	#
+	#waitpid( $chld_pid, 0 );
+	#
+	#my $child_exit_status = ($CHILD_ERROR >> 8);
+	#
+	#return $child_exit_status;
```
steveschnepp: Please *remove* the code instead of commenting it.

lib/Munin/Plugin.pm
```diff
@@ -86,7 +87,7 @@ what plugin the error message comes from.
 
 =cut
 
-our $me = (split '/', $0)[-1];
+our $me = (split '/', $PROGRAM_NAME)[-1];
```
steveschnepp: `$0` is pretty well known... actually more than `$PROGRAM_NAME` 😉

steveschnepp:
Hmm... I see the issue. It is in `.perlcritic`: `allow = $_ $!` is way too short. 

I mean, I really like `English`, but for Perl literates many don't make any sense as `$@` and `$$` are much more known than their litteral counterpart.

cgzones:
I rebased the pr and re-introduced the usage of `$$`, `$0` and `$@`.

sumpfralle:
> I mean, I really like English, but for Perl developers many do not make any sense, such as $@ and $$ which are much more known than their literal counterpart.

The last (and only) non-trivial project I implemented in perl was 15 years ago, so I am surely as far away as possible from a typical perl developer, but personally I did not know/remember `$@` and I just guessed `$$` (as in shell). Thus I would be happy about `English` without any exceptions :)
And think of all these young and fresh perl developers who blindly followed the guidance of perlcritic and maybe never ever encountered one of these `$@`?
(just sharing my opinion without putting too much weight into it)

steveschnepp:
Well... I'd argue that `$@` might be better rewrote as `try...catch` but I do see your point.

Let's settle for the current state that @cgzones kindly put together.

sumpfralle:
> Let's settle for the current state that @cgzones kindly put together.

Yes, it is a great step forward, anyway!

PR #1051, opened 2018-08-21
[perlcritic] fix some severity 4 issues

cgzones:


cgzones:
I looked over perlcritic severity 4 warnings and like to ask what to do about the following warnings:

- [Modules::ProhibitAutomaticExportation](https://metacpan.org/pod/Perl::Critic::Policy::Modules::ProhibitAutomaticExportation)

- [Objects::ProhibitIndirectSyntax](https://metacpan.org/pod/Perl::Critic::Policy::Objects::ProhibitIndirectSyntax)

- [Subroutines::ProhibitBuiltinHomonyms](https://metacpan.org/pod/Perl::Critic::Policy::Subroutines::ProhibitBuiltinHomonyms)

- [Subroutines::RequireArgUnpacking](https://metacpan.org/pod/Perl::Critic::Policy::Subroutines::RequireArgUnpacking)

- [Subroutines::RequireFinalReturn](https://metacpan.org/pod/Perl::Critic::Policy::Subroutines::RequireFinalReturn)

fix or silence them?

lib/Munin/Node/SpoolWriter.pm
```diff
@@ -16,10 +16,11 @@ use Munin::Common::Utils qw( is_valid_hostname );
 
 use Params::Validate qw( :all );
 use List::Util qw( first );
+use Readonly;
```
steveschnepp: Not sure I am a huge fan of that one as `constant` is a core feature.

script/munin-httpd
```diff
@@ -59,8 +59,6 @@ sub handle_request
 	}
 }
 
-package main;
-
 $ENV{PATH} = '/usr/bin:/bin';
```
steveschnepp: I prefer to revert to `package main;` to obviously indicate that the above code was in a package.
It could be moved to another file, or the end of this one, but it all feels not as sweet as currently.

lib/Munin/Node/SpoolWriter.pm
sumpfralle: Far from knowing anything: `Readonly` seems to be a core library and `const` is a language feature, correct?

steveschnepp: Core library might be a little stretch. It's from CPAM.

Const is from a core lib. 

Nothing is a language feature in term of const-ness.

sumpfralle: > Core library might be a little stretch. It's from CPAN.

uh - OK - my misunderstanding of a local path.

In this case I would also tend to stick to the core feature set.

cgzones:
> I looked over perlcritic severity 4 warnings and like to ask what to do about the following warnings:
    * Modules::ProhibitAutomaticExportation
    * Objects::ProhibitIndirectSyntax
    * Subroutines::ProhibitBuiltinHomonyms
    * Subroutines::RequireArgUnpacking
    * Subroutines::RequireFinalReturn
fix or silence them?

any comments?

script/munin-httpd
sumpfralle: cgzones: what do you think?

lib/Munin/Node/SpoolWriter.pm
sumpfralle: @cgzones: can we stick to `const` or is there a good reason for `Readonly`?

sumpfralle:
@steveschnepp: what do you think about cgzone's "fix or silence" question?

sumpfralle:
I am trying to clean up a bit ...

@cgzones: can we stick to `const` or is there a good reason for `Readonly`?
@steveschnepp: what do you think about https://github.com/munin-monitoring/munin/pull/1051#issuecomment-423501290?

Issue #873, opened 2017-09-12
Default port seems to be 20203


jheidbrink:
The documentation for munin-node.config states that the default value for the port directive is 4949. However, when I remove the directive from my config-file and restart the munin-node.service, munin-node reproducibly binds to port 20203. Tested with munin-node packages on Archlinux and Fedora 22.

sumpfralle:
Which version of munin-node did you install?

Did you check with your distribution (Archlinux and Fedora), if they use a rather uncommon deployment style?
Here on Debian (up to Stretch) munin-node is accessible via inetd (port 4949).

jheidbrink:
Archlinux: v2.0.26
Fedora: v2.0.33

I just took a quick look at Archlinux's PKGBUILD source for munin and couldn't see anything I would call uncommon. Also, git-grep for the port 20203 found nothing in the munin as well as in the PKGBUILD repo.

I can recreate the behaviour like this:

`systemctl stop munin-node && sed -i -e "s/^port 4949/# port 4949/" /etc/munin/munin-node.conf && systemctl start munin-node`

Which gives me the following lines in `/var/log/munin/munin-node.log`:
`2017/09/13-09:52:28 Server closing!`
`Process Backgrounded`
`2017/09/13-09:52:40 Munin::Node::Server (type Net::Server::Fork) starting! pid(32521)`
`Binding to TCP port 20203 on host X.X.X.X with IPv4`

sumpfralle:
I have found a reference to this number (20203) in the perl module [Net::Server](http://search.cpan.org/~janl/Net-Server-0.99.6.1/lib/Net/Server.pod).
I guess, that the systemd integration of munin-node caused this change (previously the "inetd" approach was common).
Did you already ask the Archlinux or Fedora maintainers?

jheidbrink:
No, but shouldn't munin-node bind to the specified default port regardless of whether it's run via inetd or standalone?

sumpfralle:
Programs that are accessible via inetd do not need to handle ports, address binding and other details of network connection. These things are defined within inetd.

It seems like your distributions decided to not use inetd for running munin-node. This is certainly fine. And maybe everything works as expected, but it just *looks* like it is wrong. You will need to discuss this with the package maintainers of munin-node in Fedora or Archlinux, if you think there is a problem.

niclan:
Hi,

There is a _significant_ startup penalty in munin-node because it runs 
[...]

PR #1056, opened 2018-08-26
munin-node: set default port if not mentioned in config file

sumpfralle:
According to the documentation, munin-node's default port is 4949.
But this default was not configured up to now.
Thus if no "port" line was specified in "munin-node.conf", then the
implicit default port of Net::Server (20203) was used.

Closes: #873

### DIFF
--- a/node/lib/Munin/Node/Server.pm
+++ b/node/lib/Munin/Node/Server.pm
@@ -153,6 +153,15 @@ sub process_request
 }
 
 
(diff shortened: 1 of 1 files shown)


sumpfralle:
The only downside of this change is, that munin-node now would switch from 20203 to 4949, if the administrator removed the `port` line from the configuration file and got used to this non-documented default port. But I do not think, that we need to take extra care for these users (if they exist).

@steveschnepp: what do you think? Merge into `stable-2.0`?

Issue #1057, opened 2018-08-26
Minor formatting error in munin-update.log (2.0.40)


kimheino:
If node is down, following entries will be written to munin-update.log:

```
2018/08/26 14:40:08 [FATAL] Socket read from cobra.bbbs.net failed.  Terminating process. at /usr/share/perl5/vendor_perl/Munin/Master/UpdateWorker.pm line 254.
2018/08/26 14:40:08 [ERROR] Munin::Master::UpdateWorker<bbbs.net;cobra.bbbs.net> died with '[FATAL] Socket read from cobra.bbbs.net failed.  Terminating process. at /usr/share/perl5/vendor_perl/Munin/Master/UpdateWorker.pm line 254.
'
2018/08/26 14:40:15 [INFO] Remaining workers: bbbs.net;cobra.bbbs.net
2018/08/26 14:40:15 [INFO] Reaping Munin::Master::UpdateWorker<bbbs.net;cobra.bbbs.net>.  Exit value/signal: 18/0
```

Note the line with only ' in it. I assume this is because ERROR-line captures and logs output, but fails to strip \n from it.

PR #1058, opened 2018-08-26
ProcessManager: remove \n from output before it's logged

kimheino:
EVAL_ERROR can be "[FATAL] Socket read from ... line 254.\n" which would
then be logged as "... line 254.\n'\n" by next line, two lines. Fix it by
removing \n and other line ends from EVAL_ERROR.

This fixes bug #1057.


sumpfralle:
Great - thank you for answering yourself :)

PR #1061, opened 2018-08-27
[munin-run] cleanup

cgzones:
- format with perltidy
- fix some perlciritc warnings


steveschnepp:
There's a lot of whitespace changes here...

cgzones:
I split it up

script/munin-run
```diff
@@ -77,78 +74,81 @@ sub main
...
 }
 
 
-sub print_version_and_exit
-{
+sub print_version_and_exit {
     require Pod::Usage;
     Pod::Usage::pod2usage(
-        -verbose => 99,
+        -verbose  => 99,
```
sumpfralle: Is this alignment of operators really a common style recommendation?
(Python's [PEP8](https://www.python.org/dev/peps/pep-0008/) defines exactly the opposite)

script/munin-run
```diff
@@ -44,31 +43,29 @@ my $paranoia = 0;
...
         defgroup   => $config->{defgroup},
         pidebug    => $PIDEBUG,
     );
 
-    $config->reinitialize({
-        %$config,
-        paranoia   => $paranoia,
-    });
+    $config->reinitialize( {
+        %{$config},    ## no critic qw(ValuesAndExpressions::ProhibitCommaSeparatedStatements)
```
sumpfralle: Just wild guessing: is this expression supposed to merge the `config` hash together with the additional hash value and supply this new hash as an argument to the `reinitialize` method?
Maybe there is a better way to achieve this without overriding the policy?

cgzones: Yeah, that would be better, but I did not yet find a way, input welcome.

script/munin-run
cgzones: It's the default perltidy setting

steveschnepp: I used to be very fond of it, as it makes the code a little more readable. But it has to be automated, which is okay with `perltidy`. 

But what I prefer nowadays is the cleanliness of diffs, and usually this adds a lot of whitespace noise. It also makes `git blame` a headache to work with. Which is a tool that I use a lot ot understand the context of when the current code was written. 

This is invaluable to be able to patch things efficiently according to the original intend. Moreover, code that I wrote 6 months ago feels just like code someone else wrote 😉 so...

script/munin-run
steveschnepp: Using a temporary var is *always* possible, and usually even preferred

script/munin-run
cgzones: Instead of `git blame` you could use [`git log -p -M --follow --stat -- path/to/your/file`](https://blog.andrewray.me/a-better-git-blame/)

steveschnepp: Err... Not really.

I don't really want to spend brain cycles on something that could be easily automated for me ;)

That said, a [git blame can be enhanced](https://jfire.io/blog/2012/03/07/code-archaeology-with-git/) to cope with it. But let's do it once for all just prior to releasing.

sumpfralle: Anyway: thank you both for sharing your hints :)

What shall we do about the alignment of operators?

steveschnepp: > What shall we do about the alignment of operators?
Let's defer all tidy changes to when we'll enforce the usage of perltidy.

cgzones:
ping

script/munin-run
sumpfralle: @cgzones: would you try to work around this (by using a separate variable?) one in order to save the warning override?

cgzones: I do not see a way to use a temporary variable :-(

sumpfralle: @steveschnepp: I think, it is time to prove your *a temporary var is always possible* claim :)

sumpfralle:
@steveschnepp: there are only a few lines of changes left in the discussion. Should we merge these?

PR #1065, opened 2018-08-28
Plugins snmp__winmem: add percentage limits

moisseev:


plugins/node.d/snmp__winmem
```diff
@@ -86,20 +100,20 @@ total.label Available memory
 total.info Total memory available
 total.draw LINE1
 EOC
+	print "used.warning ",  int($total * $ENV{warning}  / 100), "\n"
+	  if $ENV{warning};
+	print "used.critical ", int($total * $ENV{critical} / 100), "\n"
+	  if $ENV{critical};
```
sumpfralle: Here you could also use `print_thresholds("used");` (from `Munin::Plugin`), or?

See https://github.com/munin-monitoring/munin/blob/a2dec6c8afffbfa576c21c8d47036ba6ec22f5b5/lib/Munin/Plugin.pm#L373

sumpfralle: In this case you would probably also need `adjust_thresholds` for calculating absolute values based on percentage limits:

https://github.com/munin-monitoring/munin/blob/a2dec6c8afffbfa576c21c8d47036ba6ec22f5b5/lib/Munin/Plugin.pm#L389

moisseev: Thank you!
I guess, I can't use both `print_thresholds` and `adjust_threshold`, so I've chosen the latter.

plugins/node.d/snmp__winmem
```diff
@@ -86,6 +87,10 @@ my $total = $session->get_single('1.3.6.1.2.1.25.2.2.0') * 1024;
 if (defined $ARGV[0] and $ARGV[0] eq "config") {
 	my ($host) = Munin::Plugin::SNMP->config_session();
 
+my ($warning, $critical) = get_thresholds("used");
```
sumpfralle: Please take care to use the same indentation as the context.

Personally I would also prefer these initial lines (warning/critical) to be moved down to where they are used/printed. But this may be a matter of taste.

moisseev:
I could squash all this commits if you like.

sumpfralle:
> I could squash all this commits if you like.

That would indeed suit my taste, if it is no effort for you :)

Otherwise I would just merge it as it.

moisseev:
Done.

sumpfralle:
Thank you!

Issue #1066, opened 2018-09-03
Please add support for "includedir" to munin-node


sumpfralle:
Currently "munin.conf" allows to specify a directory for additional configuration files (`includedir`).

This directive could also be helpful for `munin-node` for the following use-cases:
* specify site-specific `allow` directives for the site-local munin master hosts
* specify plugins that are supposed to be ignored by `munin-node-configure` (a respective directive is not implemented, yet)

PR #1067, opened 2018-09-03
[munin-node-configure] add disable keyword to disable plugins

cgzones:
Suggested by #1066 

To disable a plugin, put a file in CONFDIR/plugin-conf.d with the content:

    [pluginname]
    disable true

### DIFF
--- a/lib/Munin/Node/Config.pm
+++ b/lib/Munin/Node/Config.pm
@@ -280,6 +280,9 @@ sub _parse_plugin_line {
     elsif ($var_name eq 'update_rate') {
         return (update_rate => $var_value);
     }
+    elsif ($var_name eq 'disable_autoconf') {
+        return (disable_autoconf => _convert_bool($var_value));
+    }
     elsif (index($var_name, 'env.') == 0) {
         return (env => { substr($var_name, length 'env.') => $var_value});
     }
@@ -337,6 +340,15 @@ sub _apply_wildcard_to_service {
 
     $self->{sconf}{$service} = $sconf;
     return;
+}
+
+
+sub _convert_bool {
+    my ($arg) = @_;
+
+    return 1 if grep(/^$arg$/x, qw(true True TRUE));
+    return 0 if grep(/^$arg$/x, qw(false False FALSE));
+    croak "Invalid boolean value '$arg'";
 }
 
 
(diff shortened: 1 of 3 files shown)


sumpfralle:
Hehe - that a tiny change, that makes a difference! :)

Initially I thought, that `munin-node.conf` would be a good place for such a setting. But your approach indeed looks preferable to me now.

For the nitpicking: I would prefer a more specific attribute like `skip_autoconf` or `disable_autoconf`.

Besides this (without looking deeper): does you change make sure, that boolean conversion is used for this attribute? (e.g. is the string `false` really interpreted as being non-true?)

lib/Munin/Node/Config.pm
```diff
@@ -340,6 +343,15 @@ sub _apply_wildcard_to_service {
 }
 
 
+sub _convert_bool {
+    my ($arg) = @_;
+
+    return 1 if grep(/^$arg$/x, qw(true True TRUE));
+    return 0 if grep(/^$arg$/x, qw(false False FALSE));
+    croak "Inavlid boolean value '$arg'";
```
steveschnepp: `Inavlid` is a typo 😉

lib/Munin/Node/Config.pm
```diff
@@ -340,6 +343,15 @@ sub _apply_wildcard_to_service {
 }
 
 
+sub _convert_bool {
```
sumpfralle: In `Munin::Common::Config` there is `_parse_bool` - maybe you could use it here?
(it accepts a broader range of string inputs)

cgzones: I didn't know about this function, but i do not like it. It's somehow confusing to me.
More acceptable values like (`on`, `off`, `1`, `0`) are feasible.

sumpfralle: Maybe you see a good way to improve the function?
I would like to prevent code duplication for boolean parsing, if possible.

steveschnepp: > More acceptable values like (on, off, 1, 0) are feasible.

We can *add* values, but removing them is always very tricky.

cgzones:
ping

sumpfralle:
Sorry for letting this one rest for so long ...

I think, this changeset is fine and would be a very good improvement.

@cgzones: could you please also add a bit of documentation for this new configuration setting?

cgzones:
added some documentation

sumpfralle:
Great - thank you!

PR #1073, opened 2018-09-17
In cdefs, replace variable names in parentheses as well

AMDmi3:
This fixes e.g. `PREV(foo)`


sumpfralle:
Great - thank you!

@steveschnepp: this should go to `stable-2.0` - or what do you think?

AMDmi3:
That would be great! Actually I've started with a patch for 2.0 (which has cdef substitution even more broken), but I've assumed this change is not suitable for merging into stable branch. See #1074

sumpfralle:
Closing this one, since #1074 was merged into `stable-2.0`. Thus this change will enter `master` quite soon, too.

PR #1074, opened 2018-09-17
Fix CDEF variable substitution

AMDmi3:
The variable name substitution in CDEFs is severely broken in munin 2.0. For instance in `foo.cdef=foo,foo,PREV(foo),+,+` only first `foo` will be replaced by corresponding `afoo` `ifoo` or `gfoo`, and rrdtool will fail to render the graph. Fix handling consecutive variable instances by using regexp lookahead/lookbehind, and extend the regexp to also handle variable name in parentheses, which will fix `PREV()`.


sumpfralle:
Thank you!

PR #1077, opened 2018-09-18
plugins/nginx_request: stats are for all ports (80+443), never for single port

kimheino:
Plugin label defaulted to text "requests port 80" and required special
config to output more modern label "requests port 443".

Nginx stats contains sum of all ports, never single port. You can not generate
stats for 80 and 443 separately. It's much simpler to just remove "env.port"
support so that port config is not needed and label is correct.


sumpfralle:
Indeed!

PR #1078, opened 2018-09-18
plugins/squid_traffic: allow negative numbers

kimheino:
Sometimes squid's status output contains negative numbers. Allow and
return them as is.


sumpfralle:
Weird ... but we need to tolerate such interesting behaviour ...

PR #1081, opened 2018-09-18
Build.PL: fix build on systems where Perl's INC path doesn't contain "."

kimheino:
This is needed on Fedora and I remember seeing something similar on Debian too.


sumpfralle:
This one should be already included (differently) since b9e6acc32c0e0856f7361f6b339815b21c6e3429.

Thank you!

PR #1084, opened 2018-09-18
master: fix warning when updating

steveschnepp:


.travis.yml
```diff
@@ -75,9 +75,7 @@ script:
   # check for potential spelling mistakes
   - codespell --quiet-level=3 --skip=ChangeLog,Announce-2.0,*.js,memory,jmx_,jmx_tomcat_dbpools,netstat_multi,munin_node_os.t
   - perl Build.PL && ./Build build && cover -test -report coveralls
-  # the old travis environment requires an old flake8 invocation
-  # TODO: fix lint issues and enable the test
-  - PYTHON_LINT_CALL="python3 -m flake8.main" make lint || true
```
sumpfralle: I guess, that we need this environment variable override :(
Probably I should have added more detail to the comment:
* the travis environment is really old and uses a packaged version of flake8, that needs to be called as done above (`python3 -m flake8.main`)
* newer packages of flake8 (i.e. at least since Debian Stretch - probably way before) use `python3 -m flake8` (and cannot handle the older style)

Thus I think it is good to remove the `|| true`, but I think you will need to stick to the `PYTHON_LINT_CALL` override. But maybe it can move somewhere to the top where environment variables are specified?

steveschnepp: Well, i just removed that one, since it is mostly handled by PerlCritic.

For testing plugins in Python, it is indeed another story.

PR #1088, opened 2018-09-21
munin-asyncd: sync munin-cron and munin-asyncd runtimes to not overlap (version 2)

kimheino:
If munin-cron and munin-asyncd are run at exactly same time for
same plugin the result is UNKNOWN status for plugin. It will change
randomly between OK and UNKNOWN, generating hundreds of warning emails
from munin-limits.

This patch adds trivial fix:

Munin-cron is run at "*/5" by cron, so run munin-asyncd
at "*/5 + 0.75 * $minrate".


sumpfralle:
Thank you for investigating the mysteries around this!

PR #1093, opened 2018-09-25
Update "special_stack" (removed in 2009) to "stack"

smokris:


sumpfralle:
Good catch - thank you!

PR #1102, opened 2018-10-10
Enable `lint-plugins` result during tests

sumpfralle:
* fix remaining style issues in a few plugins (most notable `df_abs` for Linux and many HP-UX plugins)
* enable the exitcode check of `make lint-plugins` in travis
* add missing tests for multigraph capability to some plugins
* add test for multigraph checks to the `lint-plugins` target


sumpfralle:
@steveschnepp: do you have an opinion on this change?
I think, it is quite non-controversial :)

PR #1104, opened 2018-10-12
Avoid warning with conditional instead of hiding all warnings

sumpfralle:
Get rid of one small `no warnings` block.


sumpfralle:
I assume, that `null` in this context means `undef`.

sumpfralle:
> I don't really like this style, as it changes the output.

Ah - indeed! I was mixing some python-infested thoughts into this change at that moment ...

Anyway: I have no real feelings for that change, so it shall rest in peace.

PR #1113, opened 2018-11-03
Narrow down voltage regex

RenWal:
The voltage regex (and actually all of the other regexes) doesn't look for the start of the line. This can cause UPS input voltage monitoring to fail because there are other fields that can contain the word "voltage" in the UPSC output.
Example: "input.transfer.reason: input voltage out of range" will match and cause the plugin to report
```
input.value input voltage out of range
input.value 230.0
```
This change narrows down the regex to only match at the beginning of the line.

### DIFF
--- a/plugins/node.d/nutups_.in
+++ b/plugins/node.d/nutups_.in
@@ -43,7 +43,7 @@ voltages() {
 			echo "${i}.min 0"
 		done
 	else
-		"$UPSC" "$UPS" | sed -n '/^[^ ]*volt/{
+		"$UPSC" "$UPS" | sed -n '/^[^:]*volt/{
 					s/://
 					/nominal/s/.* /nominal.value /
 					/voltage/s/\.[^ ]*/.value/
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for the explanation!

In a perfect world, you could do the following to complete this cleanup mission:
* add the anchor to the other regular expressions (where they fit)
* apply your changes to the `stable-2.0` branch (since they fix problems occurring in the stable release)

Thank you!

RenWal:
I applied the changes to all regexes as requested. However, I'm not sure how to apply them to the stable branch. Do I need to create another pull request?

sumpfralle:
I guess, that the following should work: `git rebase --fork-point HEAD~2 --onto stable-2.0`
(assuming that you want to rebase the two latest commits)
Afterwards you can force-push into the branch of this pull request.

In case something breaks you can also create a new branch from `stable-2.0` and just cherry-pick your two commits.

RenWal:
Thanks for the hint! I made the required changes, so this should be ready to pull

sumpfralle:
Great - thank you!

Issue #1100, opened 2018-10-06
smart_ broken exit status after 2.0.40 2018-08-14 merge


johnv-git:
**Describe the bug**
smart_ erroneously returns a threshold error when smartctl returns 0.  version prior to the 2.0.40 merge works as expected.  (https://raw.githubusercontent.com/munin-monitoring/munin/5619192ad8fd4d5ec3c84ca4863654a108c2a40e/plugins/node.d/smart_)

**Screenshots & Logs**
domain.com :: host.domain.com :: S.M.A.R.T values for drive ada1
    WARNINGs: smartctl_exit_status is 0.00 (outside range [1:]).
    OKs: Temperature_Internal is 100.00, Power_On_Hours is 100.00, Unsafe_Shutdown_Count is 100.00, Erase_Fail_Count is 100.00, Available_Reservd_Space is 100.00, Thermal_Throttle is 100.00, Workload_Minutes is 100.00, CRC_Error_Count is 100.00, Host_Writes_32MiB is 100.00, Workld_Media_Wear_Indic is 100.00, Workld_Host_Reads_Perc is 100.00, Current_Pending_Sector is 100.00, End_to_End_Error is 100.00, Reported_Uncorrect is 100.00, Temperature_Case is 80.00, Power_Loss_Cap_Test is 100.00, Power_Cycle_Count is 100.00, Reallocated_Sector_Ct is 100.00, Program_Fail_Count is 100.00, Media_Wearout_Indicator is 98.00, SATA_Downshift_Count is 100.00, Host_Reads_32MiB is 100.00. 

**Desktop (please complete the following information):**
 - Freebsd 11.2
 - Munin Version "2.0.42"

sumpfralle:
Could you give us a minimal example how to trigger this limits error message?
Or maybe you already suspect a specific change?

mistur:
Hello,

I confirm the patch fixed this issue (I still have the script in python2.7 of smart_, I just had to apply the patch at the right place).

Thanks,

Best regards,

PR #1117, opened 2018-11-08
Fix smartctl_exit_status (should be "1" and not "1:")

seeschloss:
Fixes bug #1100


duritong:
This fixes the problem

seeschloss:
(Okay no, sorry, I didn't see the huge diff, let me fix that later today)

seeschloss:
Should be better, I hope!

sumpfralle:
Great - thank you!

PR #1122, opened 2018-11-21
Remove tests for debian reproducible

sumpfralle:
Some of munin's tests fail within Debian reproducible test environment.
See the build log for 2.0.43-1 as an example: https://tests.reproducible-builds.org/debian/rb-pkg/unstable/i386/munin.html.

The problematic tests require network access or rely on a specific host setup. Both assumptions do not seem to be suitable for such tests.

### DIFF
--- a/node/t/munin_node_configure_hostenumeration.t
+++ b/node/t/munin_node_configure_hostenumeration.t
@@ -13,9 +13,9 @@ my @slash_24 = map { "192.168.2.$_" } 0 .. 255;
 ### resolve
 {
 	my @tests = (
-		[ 'munin-monitoring.org', 'Resolve IPv4 hostname' ],
+		[ 'localhost',            'Resolve IPv4 hostname' ],
 		[ '127.0.0.1',            'Resolve IPv4 numeric address' ],
-#		[ 'ipv6.google.com',      'Resolve IPv6 hostname' ],
+#		[ 'ip6-localhost',        'Resolve IPv6 hostname' ],
 #		[ '::1',                  'Resolve IPv6 numeric address' ],
 	);
 
--- a/node/t/munin_node_os.t
+++ b/node/t/munin_node_os.t
@@ -1,7 +1,7 @@
 use warnings;
 use strict;
 
-use Test::More tests => 30;
+use Test::More tests => 28;
 use Test::LongString;
 use Config;  # for signal numbers and names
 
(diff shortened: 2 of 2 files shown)


steveschnepp:
We really have to see how we could test it instead of simply ignoring it..

sumpfralle:
> We really have to see how we could test it instead of simply ignoring it..

Indeed the name of this merge request and my original name of one of the first commits was misleading.

The specific changes are:
* test for name resolution based on locally available data (instead of public DNS servers)
* remove the test that assumed (IMO: for no good reason) that the host has a locally configured domain (i.e. the `hostname -fdqn` returns a string containing a dot)

The DNS test should still be fine, since `Munin::Node::Configure::HostEnumeration::_resolve` uses only `gethostbyname`. Thus a test for a locally available name (e.g. via `/etc/hosts`) is just as good as a hostname resolved via a nameserver. At least this is my understanding of the purpose of this test.

The test regarding the FQDN is different: here I am not sure, what the purpose of this test is. Should it check for the support of `Sys::Hostname::hostname` on the local platform? Or should it detect a specific incompatibility of different implementations behind this function on different platforms?
The former goal could be reached by simply testing, whether the result is non-empty. The latter goal requires knowledge about the specific incompatibility, that we want to avoid.

In any case: we should really get the tests back to working again on different platforms. It would be a shame to disable the tests within the Debian packaging just for the sake of an old test code fragment of unknown origin and purpose :)

steveschnepp:
> we should really get the tests back to working again on different platforms.

Fully agreed. That said, I think we should focus more on real tests than extensive unit tests. I don't say they are not important, but the maintenance of them requires substantially more effort. At least for me.

@ssm was very proficient with those so it was a real estate. But now I'm afraid it'll slow us down. Once we'll have a test champion again in our team, we should of course reconsider.

sumpfralle:
Anyway: here we are talking about tests for stable-2.0, that were basically disabled for six years.
I do not think, we will put effort into these tests, anymore. They are just here to keep the thing working, while we are fixing small details.
Thus this merge request is not about our long-term strategy, but about our legacy test setup.

steveschnepp:
Oh. Well, I agree then ;)

That said, you should be careful when you'll merge back those commits into `master`.

sumpfralle:
> That said, you should be careful when you'll merge back those commits into `master`.

Indeed! But since you reworked the tests completely, this will be easy to spot :)

PR #1124, opened 2018-11-23
Makefile improvements

sumpfralle:
Improve some small details of the Makefiles:
* do not process Makefile.config twice
* handle fallbacks for `getent` on different platforms
* remove obsolete `Makefile.config-maint`
* handle a manually overridden Makefile location properly


steveschnepp:
Sounds good to me

PR #1126, opened 2018-12-02
Plugin memory (linux): use fixed colors for all fields

sumpfralle:
Previously only the color for the field "apps" was fixed (green). The colors of all remaining fields depended on the local setup of the host and the resulting availability of various memory-related statistics (e.g. "shmem"). A change of setup or a kernel upgrade thus lead to a confusing shift of the color palette. An unfortunate environment would even lead to the field "free" using the "red" color from the palette. This warning colour would obviously be misleading.

The memory plugin will not be maintained in the long term. It is already removed from the branch preparing the upcoming munin 3.0 release. Thus it is acceptable to introduce a fixed color scheme, even though this would usually be a maintenance burden.

Please note that the colors of almost all fields may change (once).

Definitely unchanged fields:
* apps (always the first field)
* KSM sharing (previously hard-coded)

Most other fields are *probably* unchanged for systems where the fields "page_tables", "swap_cache" and "slab_cache" are available.

Users of systems with a different set of optional memory related statistics will experience a one-time shift of the color palette for their set of available fields.

### DIFF
--- a/plugins/node.d.linux/memory.in
+++ b/plugins/node.d.linux/memory.in
@@ -102,48 +102,61 @@ if ($ARGV[0] and $ARGV[0] eq "config") {
     print "apps.label apps\n";
     print "apps.draw AREA\n";
     print "apps.info Memory used by user-space applications.\n";
+    print "apps.colour COLOUR0\n";
     print "buffers.label buffers\n";
     print "buffers.draw STACK\n";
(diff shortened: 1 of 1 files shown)


sumpfralle:
# Previous color set (for an example setup)
memory-day-old

# New color set
memory-day-new

Issue #1129, opened 2018-12-10
postgres_connections_ total count wrong with PostgreSQL 10 when using ALL config


shuehner:
Since PostgreSQL 10 pg_stat_activity view which is used as datasource for this plugin contains information about not just information about connections but other internal postgres processes.

As the plugin does not filter those new rows out it miscounts the connections by typically +5 (or more depending on postgres config (i.e. if streaming replication is used)

Documentation:
https://www.postgresql.org/docs/10/monitoring-stats.html
Table 28.3. pg_stat_activity View
backend_type column

**Type of current backend. Possible types are autovacuum launcher, autovacuum worker, background worker, background writer, client backend, checkpointer, startup, walreceiver, walsender and walwriter.**

To get accurate connection information the output needs to be filtered with: backend_type = 'client backend'.

Note:
* This only applies to the _ALL config of the plugin. If it is configured to only graph a single database this issue is indirectly avoided (as the new lines do not have database name in their output they are filtered out by the database name filter.
* As the new backend_type column was only added in PostgreSQL 10 that requires another version dependent SQL to be added  to the plugin.

**To Reproduce**
1. Configure plugin to run against a PostgreSQL 10 (or higher) database with _ALL config
2. Do not have any active connection open to any database for this PostgreSQL 10 cluster
2. munin-run postgres_connections_ALL 

**Expected behavior**
Expected output is to have 0 connections of any type

**Current, wrong behavior**
Output is:
active.value 0
idle.value 0
idletransaction.value 0
unknown.value 0
waiting.value 5

**NOTE: This issue is about the total count being 5 instead of 0. That all idle connections are reported as waiting for lock is reported as #1131

sumpfralle:
Your proposal #1137 was merged, but I noticed now, that the [documentation you referred to](https://www.postgresql.org/docs/10/monitoring-stats.html) mentions `client backend` while your commit uses  `client connection` as a filter for `backend_type`.

I am tempted to change the condition in order to match the string used in the documentation.
Or what do you think?

sumpfralle:
Ping?
(the release of 2.0.45 is approaching - I would like to finish this before)

sumpfralle:
I changed the filter to `client backend`. I guess, everything is fine now.

PR #1130, opened 2018-12-10
Fix total connection count with PostgreSQL 10 (when using _ALL config)

shuehner:
Since PostgreSQL 10 pg_stat_activity view which is used as datasource for this plugin contains information about not just information about connections but other internal postgres processes.

As the plugin does not filter those new rows out it miscounts the connections by typically +5 (or more depending on postgres config (i.e. if streaming replication is used).

Split out query to be able to have specific PostgreSQL 10 option (as new column only exists since version 10) + then add filter for backend_type = 'client connection'

Closed: #1129

### DIFF
--- a/plugins/node.d/postgres_connections_
+++ b/plugins/node.d/postgres_connections_
@@ -84,6 +84,17 @@ my $pg = Munin::Plugin::Pgsql->new(
                 ON tmp.mstate=tmp2.mstate
                 ORDER BY 1;
 		",
+            [ 9.6, "SELECT tmp.mstate AS state,COALESCE(count,0) FROM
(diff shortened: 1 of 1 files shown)


sumpfralle:
This looks good!
Thank you for taking the time to analyze the issue and document it.

I think, this improvement is suitable for the `stable-2.0` branch.
Could you please rebase your changes onto that branch?
(e.g. `git rebase --onto stable-2.0 HEAD~~`)

shuehner:
Hello @sumpfralle ,
done now. in new pull request: #1137.

Sorry probably not done nicely (but i don't know enough git).
After the rebase locally to stable-2.0 i could not push to my fork existing branch (as probably it being based on master).
So did delete+recreate it there. -> New pull request.

PR #1134, opened 2018-12-12
plugins/mailman: fix to return values if there is no rotated logfile

kimheino:


sumpfralle:
This looks good - thank you!

PR #1137, opened 2018-12-13
Fix total connection count with PostgreSQL 10 (when using _ALL config)

shuehner:
Since PostgreSQL 10 pg_stat_activity view which is used as datasource for this plugin contains information about not just information about connections but other internal postgres processes.

As the plugin does not filter those new rows out it miscounts the connections by typically +5 (or more depending on postgres config (i.e. if streaming replication is used).

Split out query to be able to have specific PostgreSQL 10 option (as new column only exists since version 10) + then add filter for backend_type = 'client connection'

Closed: #1129 

This new pull request is based on stable-2.0 branch as requested by @sumpfralle in #1130.
New pull request as i had to recreate by issue1129 after the rebase (as not knowing enough git to do that better)


sumpfralle:
This looks good to me - thank you!

PR #1140, opened 2018-12-15
Remove possible "ignore" value from "use_node_name"

sumpfralle:
The values `yes` and `ignore` for `use_node_name` have the same effect.

The `ignore` value was never documented anywhere.


sumpfralle:
This change set is based upon #1139.

sumpfralle:
> This is more to cope with none standard nodes that exist out there.

You mean: implementations of `munin-node` that cannot handle the `node_name` argument of `list`?
In this case the above change helps here, too: the `node_name` is not added as an argument anymore, if it matches the name announced by the node.

sumpfralle:
@steveschnepp: do you see any issues with the removal of this option?

PR #1143, opened 2018-12-16
plugins.linux/vlan: unify more graph_titles for network interface plu…

kimheino:
…gins


sumpfralle:
This looks good to me.

PR #1147, opened 2019-01-08
Support Samsung SSD 970 EVO

Komzpa:
Samsung SSD 970 EVO is NVMe drive. It is placed into different spot and has a bit different smartctl output layout.


sumpfralle:
That looks good to me!

Could you please apply your commit to the `stable-2.0` branch?

sumpfralle:
merged #1147 instead.
(will be merged into master soon)

PR #1148, opened 2019-01-09
Support Samsung SSD 970 EVO

Komzpa:
Samsung SSD 970 EVO is NVMe drive. It is placed into different spot and has a bit different smartctl output layout.

PR to stable-2.0 branch per request of @sumpfralle in #1147.


sumpfralle:
Thank you!

PR #1149, opened 2019-01-10
Update PostgreSQL plugin to work with Version 10 or higher

gullevek:
For postgres_connections_:
Added missing "backend_type = 'client backend'" for Version 10 or
highter

NOTE: previous it said "client connection" but this does not exist according to PostgreSQL documentation.
https://www.postgresql.org/docs/10/monitoring-stats.html#MONITORING-STATS-VIEWS
https://www.postgresql.org/docs/11/monitoring-stats.html#MONITORING-STATS-VIEWS

For perl plugin Pgsql.pm:
Updated version check to only use the major version for PostgreSQL 10
or higher

### DIFF
--- a/lib/Munin/Plugin/Pgsql.pm
+++ b/lib/Munin/Plugin/Pgsql.pm
@@ -521,7 +521,8 @@ sub get_version {
     my $v = $r->[0]->[0];
     die "Unable to detect PostgreSQL version\n"
         unless ($v =~ /^(\d+)\.(\d+).*\b/);
-    $self->{detected_version} = "$1.$2";
+    # from PostgreSQL 10 on only the major version is needed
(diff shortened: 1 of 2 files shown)


sumpfralle:
@gullevek: ping?

damluk:
```
Regarding the version comparison: you assume, that starting with postgresql 10 there will be no incompatible changes within a major version, or? Is there an announcement of that policy change public somewhere?
```
Before postgres 10, a transition from version `a.b` to version `a.c` was already a major release (https://www.postgresql.org/support/versioning/).

sumpfralle:
Versioning: thank you for clarifying that!

I just committed the changed version comparison from above in eb480e8d (thank you!)

The other part of this proposal is implemented by 6ffb285.

Issue #1144, opened 2018-12-31
Idle kernel threads missing on Linux processes


damluk:
Ubuntu disco man ps:
```
PROCESS STATE CODES
...
               D    uninterruptible sleep (usually IO)
               I    Idle kernel thread
```
This idle state is currently not graphed under Linux, and it is many of them.

sumpfralle:
In which specific plugin did you miss that information?
Maybe you want to prepare a pull request? :)

damluk:
In `plugins/node.d/processes`. Can do a merge request later.

sumpfralle:
Perfect!

damluk:
Actually it is pretty hard to be as generic as possible. According to procps changelog, Kernel 4.2 introduced the new state, so I will need to compare kernel versions. When doing that, I need to remain POSIX compliant. I won't be able to provide a generic versionstring comparison function.

sumpfralle:
How about something like the following?

```shell
[ "$min_version" = "$(printf '%s\n' "$min_version" "$current_version" | sort --version-sort)" ]
```
(or a more readable variation)

sumpfralle:
How about keeping it simple and just add the "I" indicator to the linux parser?
Older systems will see this value as zero.
Newer systems will get a real value.

What do you think?

sumpfralle:
#1157 was merged into `stable-2.0`

PR #1157, opened 2019-01-31
Graph idle kernel threads as available since Linux 4.2

damluk:
Solves #1144

### DIFF
--- a/plugins/node.d/processes.in
+++ b/plugins/node.d/processes.in
@@ -131,7 +131,7 @@ if [ "$1" = "config" ]; then
 
     # OS specific flags
     if [ "$OPERSYS" = "Linux" ]; then
-        echo "graph_order sleeping stopped zombie dead paging uninterruptible runnable processes"
+        echo "graph_order sleeping idle stopped zombie dead paging uninterruptible runnable processes"
         echo "dead.label dead"
         echo "dead.draw STACK"
         echo "dead.colour $DEAD"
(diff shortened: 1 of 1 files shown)


plugins/node.d/processes
```diff
@@ -130,7 +130,7 @@ if [ "$1" = "config" ]; then
 
     # OS specific flags
     if [ "$OPERSYS" = "Linux" ]; then
-        echo "graph_order sleeping stopped zombie dead paging uninterruptible runnable processes"
+        echo "graph_order sleeping idle stopped zombie dead paging uninterruptible runnable processes"
```
sumpfralle: I think, this would change the color of the fields in the graph (please check - maybe I am wrong).  In this case it would be great, if you moved it to the end of the list.

damluk: I don't know why it should, as the colors are defined via
```
# Define colours
RUNNABLE='22ff22'         # Green
SLEEPING='0022ff'         # Blue
STOPPED='cc0000'          # Darker red
ZOMBIE='990000'           # Darkest red
UNINTERRUPTIBLE='ffa500'  # Orange
IDLE='4169e1'             # Royal blue
PAGING='00aaaa'           # Darker turquoise
INTERRUPT='ff00ff'        # Fuchsia
LOCK='ff3333'             # Lighter red
RUNNING='00ff7f'          # Spring green
DEAD='ff0000'             # Red
SUSPENDED='ff1493'        # Deep pink
TOTAL='c0c0c0'            # Silver
```

The definition matches the resulting [color table](https://user-images.githubusercontent.com/7456652/52110155-0c50fb00-2600-11e9-82b5-270c7953a1f3.png).

sumpfralle:
> I don't know why it should, as the colors are defined via

My bad - I did not take a close look :(

Thank you!

Issue #1131, opened 2018-12-10
postgres_connections_ reports every idle connection as 'Waiting for lock'


shuehner:
**Describe the bug**
Running this plugin again PostgreSQL 10 shows every idle connection suddenly as 'waiting'.
This is related to to #746 which adjusted the plugin to use the new 'wait_event_type' column which was added in PostgreSQL 9.6

Using PostgreSQL 9.6 that already changed the behavior of the plugin to report any of the available wait_event_type as 'Waiting for lock' instead of only the type 'Waiting for lock'.
See https://www.postgresql.org/docs/9.6/monitoring-stats.html table 28-3 For list of values of wait_event_type.
Fix for #746 took any value instead of only 'Lock' and shows this as Lock to the plugin viewer.

Now in PostgreSQL 10 the list of wait_event_types was expanded on further.
https://www.postgresql.org/docs/10/monitoring-stats.html
Values here for PostgreSQL 10 include 'Client' which happens for every idle connection as it is essentially 'Waiting for the client to give it work'

That in practice makes the plugins useless as it shows all idle as 'Waiting for lock' currently.

**Steps to reproduce the behavior:**
* Configure plugin to run against a PostgreSQL 10 (or higher) database with _ALL config
* Do not have any active connection open to any database for this PostgreSQL 10 cluster
*  Create a single postgres connection i.e. by just running 'psql postgres'
* munin-run postgres_connections_ALL

**Expected behavior**
Expected output is to have 1 connection of marked as 'idle' instead of waiting.

Current, wrong behavior
Output is:
active.value 0
idle.value 0
idletransaction.value 0
unknown.value 0
waiting.value 6

NOTE: This issue is about the idle being reported as 'waiting' the total count wrong is reported in #1129 

Possible fixes could be either:
a.) Go back to pre 9.6 behavior to only report real 'Waiting for lock' which is the most frequent waiting type
b.) Rename waiting to 'Waiting to something' and only exclude 'Client' category for PostgreSQL 10
c.) Expose the full list of wait_event_types (excluding Client type)
Note: Activity type in practice also cannot happen for client connections. If #1129 is fixed that does not need to be exposed to the viewer as cannot happen.

I have patches locally for options a + c done. When direction here of the fix is decided / known i can submit those for contributions.

shuehner:
[...]

PR #1161, opened 2019-02-26
fix idle connections showing as waiting for lock in postgres >= v10

lonepsycho:
in postgresql v10 and v11 current version shows idle connections as waiting for lock. with these changes idle connections are being showed as idle, and only connections that are holding locks are shown as waiting. graph with before and after fix applied:
after_fix
Fixes: #1131


sumpfralle:

I cherry-picked your commit into `stable-2.0` as 07794c560f1a40b57109ca73a91ae811bc560708. It will reach `master` later.

Issue #460, opened 2015-04-28
Wrong links to dynamic zoom page in munin-cgi-html 


gjuric:
I understand that this option of generating graphs is deprecated but in the just released distribution of Debian (Jessie) Munin is in the latest stable version (2.0.25) which still uses this.

I have installed Munin, spawned fastgi processes for generating graphs and HTML and when I visit same pages the links to the dynamic zoom page is correct and looks like this:

```
/munin/static/dynazoom.html?cgiurl_graph=/...
```

But if I get to the same graph from the left navigation menu the link that is supposed to take me to the dynamic zoom page looks like this:

```
/munin/localdomain/localhost.localdomain/static/dynazoom.html?cgiurl_graph=/...
```

So of course, the second one does not work.

Is there an easy way to fix this?

chteuchteu:
About your second link: were you on the `/munin/localdomain/localhost.localdomain/` page?
If so, there seems to be a missing leading `/` on the dynazoom link.

We'll try to find which link to edit: once done, you'll be able to edit the templates file (`.tmpl` files used for pages generation): once the HTML files will be regenerated, the links will be fixed :)

gjuric:
Not sure anymore, I tried it again and this navigation path does not work:

    /munin/
    /munin/localdomain/localhost.localdomain/index.html#disk
    /munin/localdomain/localhost.localdomain/diskstats_latency/index.html
    /munin/localdomain/localhost.localdomain/diskstats_latency/sda.html
    /munin/localdomain/localhost.localdomain/diskstats_latency/static/dynazoom.html?cgiurl_graph=/cgi-bin/munin-cgi-graph&plugin_name=localdomain/localhost.localdomain/diskstats_latency/    sda&size_x=800&size_y=400&start_epoch=1430203806&stop_epoch=1430311806

This one works ok:

    /munin/
    /munin/localdomain/localhost.localdomain/index.html#network
    /munin/localdomain/localhost.localdomain/fw_conntrack.html
    /munin/static/dynazoom.html?cgiurl_graph=/cgi-bin/munin-cgi-graph&plugin_name=localdomain/localhost.localdomain/fw_conntrack&size_x=800&size_y=400&start_epoch=1430203806&stop_epoch=1430311806

I am using CGI strategy for generating graphs as I mentioned in the first post. Not sure if that has an impact.

chteuchteu:
Could you try to edit the `munin-serviceview.tmpl` template in `/etc/munin/templates`?
[...]

PR #1163, opened 2019-03-01
munin-cgi-html: avoid accidental permanent path manipulation

sumpfralle:
In case munin-cgi-html is exeucuted in a permanent context (e.g. via
fcgid or as a systemd socket service) it accidentally manipulated a path
variable permanently.

The issue can be reproduced with the following steps:
* request a service page (e.g. "/munin/GROUP/HOST/df.html")
    * links to "dynazoom" are valid
* request a time-based category comparison (e.g. "/munin/disk-day.html")
* request a service page (e.g. "/munin/GROUP/HOST/df.html")
    * links to "dynazoom" lack the relative "../../" prefix

Thus we need to restore the previous path settings after emitting a
top-level page (e.g. a time-based category comparison).

This fix is a bit hacky (just like the problematic path manipulation
itself).  But given that `munin-2.0` is effectively in maintenance-only
mode, this rather cumbersome minimal fix should be acceptable.

Closes: #460

Thanks to gnuheidix (https://github.com/gnuheidix) for preparing an
excellent minimal procedure for reproducing this issue.

### DIFF
--- a/master/lib/Munin/Master/HTMLOld.pm
+++ b/master/lib/Munin/Master/HTMLOld.pm
@@ -285,6 +285,9 @@ sub emit_comparison_template {
     # or even worse within a category inside it. We strip out the
     # extra '../' that is used to generate a relative path which is no
     # longer valid.
+    # Sadly this change is permanent and will influence successive
+    # responses in a permanent execution environment like fcgid or
+    # as a systemd socket service.  Thus we need to revert it later.
     foreach my $cat(@{$key->{'comparecategories'}}) {
         foreach my $service(@{$cat->{'services'}}) {
             foreach my $node(@{$service->{'nodes'}}) {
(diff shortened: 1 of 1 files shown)


sumpfralle:
Just for the record: the proper fix would be a reworked path handling (instead of one-shot manipulations for specific situations) in `munin-cgi-html`. But we surely do not want to do this (and it is not worth the effort), since `munin-2.0` is in maintenance-only mode.

gnuheidix:
I've successfully tested this change in my CGI based setup which I've used in #460.

I watched all logs while testing and I didn't find any new abnormalities. I recommend that somebody without the CGI stuff tests it as well to make sure we didn't add any regressions in that mode of operation.

I'm very happy that we've nailed this bug. :+1:

sumpfralle:
I also did not notice any issues in my CGI based setup (which did not expose the issue before).

Thus it should be working.

PR #1167, opened 2019-03-11
Sanitise service names

ProgVal:
When using munin-async, plugins are seen by the master as services (the plugin name being `__root__`), so they were not escaped as plugin names are.
This causes broken links, because they are unsanitised, while sanitised names are used when opening HTML and image files.

Possibly related: http://munin-monitoring.org/ticket/1725

---

I did not find how to run munin's automated tests. I'll look into it if needed.

### DIFF
--- a/master/lib/Munin/Master/Node.pm
+++ b/master/lib/Munin/Master/Node.pm
@@ -329,7 +329,7 @@ sub parse_service_config {
 	if ($line =~ m{\A multigraph \s+ (.+) }xms) {
 	    push_graphorder($service);
 
-	    $service = $1;
+	    $service = $self->_sanitise_plugin_name($1);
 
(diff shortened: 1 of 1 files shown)


ProgVal:
nevermind, it actually breaks more stuff

sumpfralle:
But the issue you encountered still exists, or?

Do you have an idea how we could fix it?

ProgVal:
Yes, the issue still exists, but I don't know how to fix it.

ProgVal:
A possible fix is to sanitise paths in `munin_get_picture_loc`. But given how the code is structured, it looks like sanitisation should happen sooner in the pipeline. I'm still looking into it.

sumpfralle:
> I'm still looking into it.

Cool!

PR #1170, opened 2019-03-16
Sanitise plugin names in munin-asyncd.

ProgVal:
Follow-up on #1167

Related:
* http://munin-monitoring.org/ticket/1725
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=710529

### DIFF
--- a/node/_bin/munin-asyncd.in
+++ b/node/_bin/munin-asyncd.in
@@ -125,7 +125,10 @@ my @plugins;
 			"w",
 		);
 
-		print $fh_list $plugins_line;
+		my $sanitised_plugins_line = $plugins_line;
+		$sanitised_plugins_line =~ s/[^_A-Za-z0-9 ]/_/g;
+
+		print $fh_list $sanitised_plugins_line;
 		print $fh_list "\n";
 	}
 
--- a/node/lib/Munin/Node/SpoolWriter.pm
+++ b/node/lib/Munin/Node/SpoolWriter.pm
@@ -102,6 +102,10 @@ sub write
 {
     my ($self, $timestamp, $service, $data) = @_;
 
+    # squash the $service name with the same rules as the munin-update when using plain TCP
+    # Closes D:710529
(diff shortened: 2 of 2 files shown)


sumpfralle:
Thank you for preparing this patch!

It looks good to me. 
But since I am not using the `munin-async` transport at the moment, I cannot really test this change.

@ProgVal: how is your test setup running? Did you notice any issues?

@steveschnepp: maybe you want to comment this change, too?

ProgVal:
> @ProgVal: how is your test setup running? Did you notice any issues?

I'm not sure I understand the question. I have a "production" munin that had issues with plugins whose name contain a dash (eg. `if_` with interfaces containing a dash, or `libvirt-mem`) -- file names contained an underscore while links contained a dash. And this patch fixed the issue without creating new ones afaik.

I don't really understand how munin uses unit tests, do you want me to write one?

sumpfralle:
@ProgVal: thank you for confirming, that your setup works with the above changes. As I said: I do not use the async protocol, thus I can only infer from the code changes, whether they have the desired effects.
Following your suggestion I added some changes to the existing unittests of munin. Please note, that `stable-2.0` is in maintenance-only mode - thus the changes are rather minimal (instead of adding separate tests for the name munging).

Please take a look at my test changes (see #1185) and also check, whether my summary of these changes (in the description of that pull request) matches your understanding of the changes you proposed.

Please understand, that I am a bit cautious here, since I do not use this part of munin, but I want the upcoming release of munin (which will most likely be the final one to be included in the upcoming Debian Buster release) to be in good shape.

Thank you!

ProgVal:
> Please take a look at my test changes (see #1185) and also check, whether my summary of these changes (in the description of that pull request) matches your understanding of the changes you proposed.

The changes and description look good, thanks.

btw, thanks for working on munin, I really like it.

sumpfralle:
Your commit is merged via #1185.

Thank you for discussing and verifying the changes in depth!

Issue #579, opened 2015-10-13
Host disappears from Munin::Overview index.html in case of failed connection to host


myroslav:
This is essentially reopening of http://munin-monitoring.org/ticket/758.

> I just noticed, that if a host which is monitored fine by munin master, stops responding (for example, munin-node is down, or no route to host), it disappears from the Munin::Overview html page.

This is still happening with 2.0.4 and 2.1.2

ssm:
The "munin-update" process creates its state file each run.  This state file is used by other parts of munin, for instance to generate html.

This is also visible if you run "munin-update --host something.example.com", as everything else will disappear from the overview until munin-update has run again.

mrch25:
I have the same issue in:
* Munin 2.0.19-3
* Ubuntu 14.04.2 LTS

ssm:
This affects all known versions of munin, until munin-update is rewritten to use a persistent database.

ssm:
Related feature request: #720

sumpfralle:
[niclan commented](https://github.com/munin-monitoring/munin/issues/1091#issuecomment-424438354) in a duplicate issue (#1091):

> In 2.0 there used to be some code to copy out enough of the data tree to preserve the html so the graphs would be viewed even if the node went temporarily offline. I think this code was supposed to keep the node in the html indefinitely - as long as the node is configured. But even in 2.0 I saw that nodes dropped out while still configured - but I never delved into why.

sumpfralle:
I think, I found a (or maybe *the*) cause for such losses of previous history.

Consider the following scenario:
1. A plugin breaks somehow and emits trash to its stdout.
2. The malformed content [causes a warning but is stored anyway in the datafile](https://github.com/munin-monitoring/munin/blob/stable-2.0/master/lib/Munin/Master/Utils.pm#L689). This tolerance may be a sane approach, in order to support future changes. I cannot judge.
3. When the `datafile` is parsed, any parsing error (e.g. for the malformed line above) is allowed to rise through the caller chain up to file parsing itself (e.g. see [`_parse_config_line` tolerating a failure](https://github.com/munin-monitoring/munin/blob/stable-2.0/master/lib/Munin/Master/Config.pm#L475)).

Thus a single malformed line in `datafile` breaks its parsing. Thus old config data is always lost in the next run. This causes nodes to disappear immediately from the HTML representation.

I am working on a fix.

sumpfralle:
A small addition to my previous description:
[...]

PR #1176, opened 2019-04-07
Ignore malformed lines when parsing "datafile"

sumpfralle:
Previously a single malformed line in "datafile" was treated as a file
parsing error.  All subsequent lines were ignored in this case.
This could cause the immediate removal of unreachable hosts from the
HTML listing, if their old configuration lines were lost due to the
above behaviour.

Now the "parse_config" subroutine accepts an optional
"skip_line_on_error" parameter. It is enabled when parsing "datafile"
and stays disabled for all other configuration files.

Closes: #579

### DIFF
--- a/master/lib/Munin/Master/Config.pm
+++ b/master/lib/Munin/Master/Config.pm
@@ -470,17 +470,26 @@ sub _split_config_line_ok {
 
 sub _parse_config_line {
     # Parse and save contents of random user configuration.
-    my ($self, $prefix, $key, $value) = @_;
-    
-    my $longkey = $self->_concat_config_line_ok($prefix,$key,$value);
+    my ($self, $prefix, $key, $value, $skip_on_error) = @_;
+    my $longkey;
 
+    if ($skip_on_error) {
+        eval { $longkey = $self->_concat_config_line_ok($prefix, $key, $value); };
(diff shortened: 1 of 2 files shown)


sumpfralle:
@steveschnepp: could you please verify, that my change is proper perl style? The logic should be fine - I am just not sure about the stylish correctness of the optional argument for the `parse_config` subroutine.

master/lib/Munin/Master/Config.pm
```diff
@@ -470,17 +470,26 @@ sub _split_config_line_ok {
...
 
+    if ($skip_on_error) {
+        eval { $longkey = $self->_concat_config_line_ok($prefix, $key, $value); };
+        # Skip single malformed lines (error messages were emitted before).
+        # This is suitable for fault tolerant parsing of "datafile".
+        return if $EVAL_ERROR;
+    } else {
+        # A problem with a malformed line is allowed to rise to the file level.
+        # This is suitable for configuration files.
+        $longkey = $self->_concat_config_line_ok($prefix, $key, $value);
```
steveschnepp: You could always evaluate, and defer only the error handling depending on the `$skip_on_error`, but I don't think it's worth changing the code, as it will be removed anyway in 3.0.

And, right now, it's dead-simple to read & already working, so let's keep it that way.

PR #1182, opened 2019-04-25
Add note about field name parsing

stijn-uva:
Numbers at the start of field names are interpreted as underscores internally 
(or by RRDTool?) This was not documented, but can be problematic, and thus
this adds a note for this in the docs.


sumpfralle:
Thank you!

Issue #1179, opened 2019-04-18
Setting tls_private_key/tls_certificate per node has no effect


neirbowj:
**Describe the bug**

The master always tries to load its certificate and key from the locations defined in the global section of the configuration, even when overridden on a per node basis.

**To Reproduce**

1. Add a node using paranoid mode
```
# munin.conf
[Group;host.example.com]
    tls paranoid
    tls_certificate /non/default/location/cert.pem
    tls_private_key /non/default/location/key.pem
    address host.example.com
```
2. Create the key and certificate in the specified location
3. Run munin-update

**Expected behavior**
`tls_private_key` and `tls_certificate`in the per-node config takes precedence over the default/global values.

**Screenshots & Logs**
```
2019/04/18 16:00:19 [WARNING] No key file "/etc/munin/munin.pem". Continuing without private key.
2019/04/18 16:00:19 [WARNING] No certificate file "/etc/munin/munin.pem". Continuing without certificate.
````

**Desktop (please complete the following information):**
 - OS+Distribution Version: RHEL 6.10
 - Munin Version: munin-2.0.43-1.el6.noarch

sumpfralle:
@neirbowj: could you please verify, whether #1184 would solve this issue for you?

neirbowj:
@sumpfralle Yes, this would solve it. Thank you very much.

sumpfralle:
Fixed in 574d102b322b23da72e687c7cbe130d28f8fa8e2.

PR #1184, opened 2019-04-27
Allow node-specific TLS settings on the master

sumpfralle:
Some masters may want to use different certificates or certificate
authorities for different nodes.

All settings starting with `tls` are parsed from the global master configuration, as well as from the current node section.

Closes: #1179

### DIFF
--- a/master/lib/Munin/Master/Node.pm
+++ b/master/lib/Munin/Master/Node.pm
@@ -151,13 +151,19 @@ sub _do_connect {
     return 1;
 }
 
+
+sub _get_node_or_global_setting {
+    my ($self, $key) = @_;
+    return exists $self->{configref}->{$key} ? $self->{configref}->{$key} : $config->{$key};
+}
+
+
 sub _run_starttls_if_required {
     my ($self) = @_;
 
     # TLS should only be attempted if explicitly enabled. The default
     # value is therefore "disabled" (and not "auto" as before).
-    my $tls_requirement = exists $self->{configref}->{tls} ?
-                                   $self->{configref}->{tls} : $config->{tls};
+    my $tls_requirement = $self->_get_node_or_global_setting("tls");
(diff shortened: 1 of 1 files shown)


sumpfralle:
@steveschnepp: would you consider this to be an acceptable change for munin 2.0? (see #1179)

master/lib/Munin/Master/Node.pm
```diff
@@ -151,13 +151,19 @@ sub _do_connect {
     return 1;
 }
 
+
+sub _get_node_or_global_setting {
+    my ($self, $key) = @_;
+    return exists $self->{configref}->{$key} ? $self->{configref}->{$key} : $config->{$key};
```
steveschnepp: using `defined()` is usually better than `exists`.

But, I agree... here the original line was using `exists`, so it might be safer for _this_ particular change.

sumpfralle:
> But I'm not at ease to let it into 2.0, as it feels more like a feature than a bug fix.

I took a look at the [documentation](http://guide.munin-monitoring.org/en/latest/advanced/tls-setup.html?highlight=tls#selective-tls). At the moment it mentions only the following:

> From the source code, it seems you can even use different certificates for different hosts. This, however, has not been tested for the purpose of this article.

Thus indeed the change would not really be a bug-fix for a documented feature. But one could argue, that the current implementation lacks the obvious "separate tls setup per peer" feature.

> Moreover, we don't have good testing coverage for that part, so that adds to my uneasiness.

Indeed. On the other hand: the change is quite simple and it is hard to imagine, how it could break existing setups.

@steveschnepp: I am undecided - both ways (merging or abandoning) are acceptable for me. Please throw a coin (or reason about your preferred choice) and pick one of them.

Thank you for the hint with "exists vs. defined"!

steveschnepp:
let's merge it...

PR #1189, opened 2019-05-22
Add plugin for PostgreSQL checksum failures

jeltz:
PostgreSQL 12 adds a per database counter for data checksum failures
which is something which is interesting for Munin to monitor. The
plugin has its critical thresholds configured to 0:0 by default since
any checksum failure is almost guaranteed to be bad news.

PostgreSQL 12 beta1 will soon be released and PostgreSQL 12 itself is scheduled to be released this autumn.


sumpfralle:
That looks good to me - thank you!

PR #1203, opened 2019-06-14
doc: fix example apache virtualhost RewriteRule

kenyon:
I just recently built a new munin master with 2.0.49 on Debian with Apache, and found that the example was wrong here. Not sure why it was changed like this in commit fa0dfe596afcc4dfef9168e15e1f40d70e7eec4f.


sumpfralle:
That looks reasonable - thank you!

PR #1205, opened 2019-06-20
Don't "exit 1"

niclan:
Postfix does not make the log file until there is anything to log.  So on low-volume hosts this makes unneeded noise.

### DIFF
--- a/plugins/node.d/postfix_mailvolume
+++ b/plugins/node.d/postfix_mailvolume
@@ -145,7 +145,7 @@ my $logfile = "$LOGDIR/$LOGFILE";
 
 if (! -f $logfile) {
     print "volume.value U\n";
-    exit 1;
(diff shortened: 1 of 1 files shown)


sumpfralle:
Great!

Would you mind to apply this change to `stable-2.0` instead?

sumpfralle:
@steveschnepp: do you consider this change to be unsuitable for `stable-2.0`? Or did you just want to relieve niclan from rebasing the commit?
(I am fine with both - just curious)

steveschnepp:
I did feel that it is more an enhancement than a bugfix.
Yet as it's mostly harmless, feel free to cherry-pick it into `stable-2.0` if you think it would be a good idea.

sumpfralle:
OK - I did that now.

Thanks for your explanation!

PR #1213, opened 2019-08-05
Fix default port in install.rst.

mstenta:
I was installing Munin today based on the instructions here: http://guide.munin-monitoring.org/en/latest/installation/install.html

The docs say that `munin-http` will be running on port 4948, but it is running on 4949 on my system. Maybe just a typo in the docs? Or is there something else I'm unaware of that explains it?

### DIFF
--- a/doc/installation/install.rst
+++ b/doc/installation/install.rst
@@ -13,7 +13,7 @@ collect data from all nodes, and graph the results. When starting with
 Munin, it should be enough to install the Munin master on one server.
 
 The munin master runs :ref:`munin-httpd` which is a basic webserver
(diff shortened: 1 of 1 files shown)


sumpfralle:
> The docs say that munin-http will be running on port 4948, but it is running on 4949 on my system.
> Maybe just a typo in the docs? Or is there something else I'm unaware of that explains it?

`munin-httpd` (available since munin 2.999) should really run on 4948 . The process you see on 4949 is `munin-node`.

If you have an idea, how to avoid misunderstandings in the documentation: please share your thoughts (or proposals)!

mstenta:
Ah my mistake - thanks for the clarification!

PR #1216, opened 2019-08-09
Update apt_all

mafri27:
"last" was triggered at each iteration of the @releases array. So the print was only executed if current line of state file is the first entry of @releases.


sumpfralle:
Merged into `stable-2.0` via #1217. `stable-2.0` is merged into `master` from time to time.

PR #1218, opened 2019-08-10
create plugin to monitor icinga2 state

mafri27:
create plugin to monitor icinga2 state


plugins/node.d/icinga_hosts
```diff
@@ -0,0 +1,51 @@
...
+##%# family=auto
+
+#%# capabilities=autoconf 
+
+if [ "$1" = "autoconf" ]; then
+    if [ -r /usr/bin/icingacli ]; then
+        echo yes
+        exit 0
+    else
+        echo "no (could not found icingacli)"
```
sumpfralle: I would recommend to use something like the following in the header of the script:
```shell
ICINGACLI=${ICINGACLI:-$(which icingacli)}

Later you can check, whether "$ICINGACLI" is empty or the referenced exists.
This allows to use the plugin in different environments (with different paths) simply by supplying an adjusted default configuration (without changing the code of the plugin).
```

plugins/node.d/icinga_hosts
```diff
@@ -0,0 +1,51 @@
+#!/bin/sh
+
+##%# family=auto
```
sumpfralle: Please add a perldoc header here (see for example the `plugins/node.d.linux/if_` plugin).

plugins/node.d/icinga_hosts
```diff
@@ -0,0 +1,51 @@
...
+        echo "no (could not found icingacli)"
+        exit 1
+    fi
+fi
+
+export TERM=vt100
+
+if [ "$1" = "config" ]; then
+
+    echo "graph_title Icinga Hosts"
```
sumpfralle: Maybe "Icinga host checks" instead?

plugins/node.d/icinga_hosts
```diff
@@ -0,0 +1,51 @@
...
+    echo "up.draw AREA"    
+    echo "down.draw STACK"
+    echo "unreachable.draw STACK"
+    echo "pending.draw STACK"
+
+    exit 0
+fi
+
+
+output=$(icingacli monitoring list hosts)
```
sumpfralle: I would recommend to the json output format combined with a filtering program instead:
```shell
data=$(icingacli monitoring list hosts --format=json)
host_up=$(echo "$data" | jq -r '.[] | select(.host_state != 0) | .host_name' | wc -l)
```

`jq` is used by multiple plugins. Thus it is quite common to depend upon it (to be mentioned in the documentation header and to be checked in the `autoconf` test).

mafri27:
I've done some changes that are hopefully cover your notes (expect the perldoc header because I've never done this before).

But If this request is at the wrong repository we can close this request and I will try it on the rigth one again. 

perldoc header

plugins/node.d/icinga
```diff
@@ -0,0 +1,82 @@
...
+##%# family=auto
+
+#%# capabilities=autoconf
+
+ICINGACLI=${ICINGACLI:-$(which icingacli)}
+JQ=${JQ:-$(which jq)}
+
+
+if [ "$1" = "autoconf" ] ; then
+    if [ "$ICINGACLI" != ""  -a "$JQ" != "" ] ; then
```
sumpfralle: Here I have two suggestions:
* it is quite common to use `[ -z "$foo" ]` for testing an empty string
* `-a` (or `-o`) are [not portable and should be avoided](http://mywiki.wooledge.org/BashPitfalls#A.5B_.22.24foo.22_.3D_bar_.26.26_.22.24bar.22_.3D_foo_.5D)

Thus maybe you want to switch to the following style for a more precise error message?
```shell
if [ ! -x "$ICINGACLI" ]; then
    echo "no (could not find 'icingacli')"
elif [ ! -x "$JQ" ]; then
    echo "no (could not find 'jq')"
else
    echo "yes"
fi
```

plugins/node.d/icinga
```diff
@@ -0,0 +1,82 @@
...
+        echo yes
+        exit 0
+    else
+        echo "no (could not found icingacli or jq)"
+        exit 1
```
sumpfralle: "autoconf" should *always* exit with zero. A non-zero exitcode indicates a plugin failure (not a missing requirement).

plugins/node.d/icinga
```diff
@@ -0,0 +1,82 @@
...
+    fi
+fi
+export TERM=vt100
```
sumpfralle: Is this really necessary? If yes: please add a comment explaining it.
(otherwise we could never get rid of this line)

plugins/node.d/icinga
```diff
@@ -0,0 +1,82 @@
...
+
+if [ "$1" = "config" ]; then
+
+    echo "multigraph icinga_hosts"
```
sumpfralle: I think `icinga_host_checks` would be a bit more specific. Maybe in the future the plugin will also support "icinga hosts" in a cluster (or something else host-related) ...
(but this is a matter of taste)

plugins/node.d/icinga
```diff
@@ -0,0 +1,82 @@
...
+    echo "ok.draw AREA"
+    echo "warning.draw STACK"
+    echo "critical.draw STACK"
+    echo "unknown.draw STACK"
+    echo "pending.draw STACK"
+
+    exit
+fi
+
+output=$($ICINGACLI monitoring list hosts --format=json)
```
sumpfralle: Please quote variables (here: `$ICINGACLI`) in order to avoid whitespace issues.
`shellcheck` is your friend! :)

plugins/node.d/icinga
mafri27: I've changed this. Reason was that I had some problems with the which output.

plugins/node.d/icinga
mafri27: also changed.

plugins/node.d/icinga
mafri27: this was caused by some special characters at the text-output of icingacli. This is not needed after the change to json,

plugins/node.d/icinga
mafri27: is renamed

plugins/node.d/icinga
mafri27: this is also done

plugins/node.d/icinga_checks
```diff
@@ -0,0 +1,113 @@
...
+    echo "warning.draw STACK"
+    echo "critical.draw STACK"
+    echo "unknown.draw STACK"
+    echo "pending.draw STACK"
+
+    exit
+fi
+
+if [ ! -x "$ICINGACLI" ]; then
+    echo "could not find 'icingacli'"
```
sumpfralle: Please output these error messages to stderr (`>&2`). Otherwise you would confuse the output parser.

mafri27:
i've changed this and opened a request at `contrib:master`. Thank you for your help.

PR #1221, opened 2019-08-28
snmp__df_ram: handle undef and zero values

moisseev:


plugins/node.d/snmp__df_ram
```diff
@@ -198,7 +198,8 @@ foreach my $storage (keys %$stor_id)
 
 foreach my $part (keys %partitions)
 {
-    print (&get_name_by_mp ($part), ".value ", ($partitions{$part}{used}*100/$partitions{$part}{size}), "\n");
+    print &get_name_by_mp ($part), ".value ", ($partitions{$part}{used}*100/$partitions{$part}{size}), "\n"
+      if (defined $partitions{$part}{used} && $partitions{$part}{size});
```
sumpfralle: Thank you for your suggestion!

Did you mean `defined $partitions{$part}{used} && defined $partitions{$part}{size}`?

(my perl is not too great - thus maybe you know better ...)

moisseev: No, I did that on purpose. In both cases we need to avoid using of uninitialized values in arithmetic operations. In case of `size` we need to avoid division by zero as well, whereas zero is a valid value for `used`.

sumpfralle: The wonders of perl's implicit expressions ...

Thanks for your explanation and contribution!

PR #1223, opened 2019-09-06
snmp__if_: handle undefined values

moisseev:
when the host is down


moisseev:
ping

sumpfralle:
Sorry - I overlooked your change. Thank you for the reminder and the fix!

PR #1224, opened 2019-09-06
snmp__processes: handle undefined values

moisseev:
when the host is down

### DIFF
--- a/plugins/node.d/snmp__processes.in
+++ b/plugins/node.d/snmp__processes.in
@@ -93,5 +93,5 @@ EOC
 }
 
 my $session = Munin::Plugin::SNMP->session();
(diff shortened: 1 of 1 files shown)


moisseev:
> *Could you please change your commit in order to print the magic value `U` ("unavailable") instead of omitting the `value` line?

Done. Could you please explain why is it important? I tried it both ways (`U` and omiting the line) and found no difference in Munin behavior.

sumpfralle:
> Done.

Thanks!

> Could you please explain why is it important? I tried it both ways (U and omiting the line) and found no difference in Munin behavior.

Munin emits warning messages for missing `value` lines. This helps catching errors/typos in plugins. Additionally munin's `limit` handling treats them differently.

PR #1226, opened 2019-09-10
[plugins/ping_] Add support for ping4

shtrom:
ping(8) from iputils handles both IPv4 and IPv6 when called as `ping`.
It is however possible to pin the address family with symlinks named
`ping4` and `ping6`.

This commit introduces support for \`ping4\` to force IPv4 pings with
iputils.


sumpfralle:
Looks good to me - thank you!

PR #1231, opened 2019-09-29
Name in title master

shtrom:
Expose NAME in all services HTML templates, and  put it first in <title>.

Useful when many tabs are open.


sumpfralle:
Could you please give an example of the previous and the future title to be used?

shtrom:
Type | Old | New
-------|------|-------
Node | `Munin Munin :: domain :: host` | `host ( Munin :: domain :: host)`
Service | `Munin Munin :: domain :: host :: service` | `service ( Munin :: domain :: host :: service)`
Screenshot| Screenshot from 2019-09-30 20-29-35 | Screenshot from 2019-09-30 20-29-39

sumpfralle:
@shtrom: thank you for this great illustration!

@steveschnepp: do you have an opinion?

I would tend to agree, that the proposed style fits better to the currently common visualization in browsers.

Details:
* the whitespace within the brackets was probably not on purpose
* we can probably get rid of the final component in brackets? (`service` and `host`)

shtrom:
On Mon 30 Sep 2019 at 08:03:16 -0700, Lars Kruse wrote:
>Details:
>* the whitespace within the brackets was probably not on purpose

Ah, yes.

>* we can probably get rid of the final component in brackets? (`service` and `host`)

Difficult. The template is looping through a list that is built by the 
Perl code. We could remove the last element of the array there, but I 
wouldn't want to create trickle-down unintended effects to templates 
that may be relying on it being there.

-- 
Olivier Mehani <shtrom@ssji.net>
PGP fingerprint: 4435 CF6A 7C8D DD9B E2DE  F5F9 F012 A6E2 98C6 6655
Confidentiality cannot be guaranteed on emails sent or received unencrypted.

sumpfralle:
@steveschnepp: do you have an opinion?
Otherwise I would merge it.

sumpfralle:
No further comments arrived - thus I would merge it now. Sorry for the long delay :(

@shtrom: could you check please, whether this proposal (#1231), #1232 and https://github.com/munin-monitoring/contrib/pull/1019 are consistent? Thank you!

shtrom:
Yup, I've been running on them for a while. All good to go :+1:

sumpfralle:
I merged them all now.
Thank you!

PR #1233, opened 2019-09-30
[snmp] Update some titles, categories and draw to better match other …

shtrom:
…plugins

This allows for better grouping, and more unified look, of graphs
showing the same metric collected by either NMP and non-SNMP plugins
across different hosts.


sumpfralle:
Thank you!

Issue #874, opened 2017-09-12
Dynazoom: dates replaced with "NaN" on Safari / Mac OS


stephane-isvtec:
On Safari / Mac OS when updating the dates values (either start_iso8601, stop_iso8601, start_epoch or stop_epoch) in dynazoom and validating using the "update" button, dates are replaced with "NaN" values.

Url example:
- before: https://xxx.xxx/munin/static/dynazoom.html?cgiurl_graph=/munin-cgi/munin-cgi-graph&plugin_name=Default/HOSTNAME/load&size_x=800&size_y=400&start_epoch=1505094760&stop_epoch=1505202760
- after update: https://xxx.xxx/munin/static/dynazoom.html?plugin_name=Default%2FHOSTNAME%2Fload&start_iso8601=2017-09-11T05%3A52%3A40%2B0400&stop_iso8601=2017-09-12T11%3A52%3A40%2B0400&start_epoch=NaN&stop_epoch=NaN&lower_limit=&upper_limit=&size_x=800&size_y=400&cgiurl_graph=%2Fmunin-cgi%2Fmunin-cgi-graph

![screen shot 2017-09-12 at 16 15 39](https://user-images.githubusercontent.com/22585734/30325408-abcc8c72-97d5-11e7-96d5-6833d2bd4dbe.png)

GoodBoy962:
Could you give a plan how to install munin on Mac?

dipohl:
On Sat, 23 Sep 2017 16:19:51 +0000 (UTC)
Pliskin Alexander wrote:

> Could you give a plan how to install munin on Mac?
> 

Which version of Munin do you want to install?

There exists a (very old) tutorial in Trac wiki:
http://munin-monitoring.org/wiki/MuninInstallationDarwin

Concerning the webserver configuration
you can find some examples in the Munin Guide:
http://guide.munin-monitoring.org/en/latest/example/index.html

HTH,

Gabriele

GoodBoy962:
@dipohl I wont install it on mac os. I used this manual for that https://codingnightly.wordpress.com/2013/10/27/monitor-macosx-using-munin/. But when I open the url in the browser there is just only a link to the parent folder. And there is nothing in /opt/local/www/munin folder

dipohl:
On Sun, 24 Sep 2017 19:40:04 +0000 (UTC)
Pliskin Alexander wrote:

[...]

PR #1245, opened 2019-10-28
Make date parsing work in Safari

cycloon:
Safari does not support `new Date()` using that old string format. To make this work across all browsers, use `new Date()` using single parts of a date.

Started from #874


sumpfralle:
Thank you!

PR #1246, opened 2019-11-01
Plugin sensors_: add support for unit prefix

sumpfralle:
Adjust the parsing of the `sensors_` plugin for a new output format.

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=943577


sumpfralle:
Merged as e726812574af01ffb13ce1c7c2a391e2eb38a71a.

PR #1248, opened 2019-11-13
RFC: add minimal local plugin management tool

sumpfralle:
The script allows to download, install and update plugins from a plugin repository (e.g. munin's contrib repository).

The name (`munin-get`) and the concept was mentioned in a [munin IRC meeting some years ago]( http://meetbot.debian.net/munin/2015/munin.2015-04-15-19.31.log.html).

The implementation is a shell script (shellcheck-clean and well documented). I do not expect it to grow beyound the point of reasonable maintainability.

What do you think about this approach?
How do you like the name (`munin-get`) - maybe you have other ideas?
Should such a script be included in munin releases?

### DIFF
--- a/script/munin-get
+++ b/script/munin-get
@@ -1,0 +1,612 @@
+#!/bin/sh
+
+: << =cut
+
+=head1 NAME
+
+munin-get - a simple tool for the management of plugins from remote git repositories
+
+
+=head1 FEATURES
+
+The small management tool "munin-get" supports the following actions:
+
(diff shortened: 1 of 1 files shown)


kenyon:
I see that it has perldoc, but would it have a man page as well? Also, some documentation for https://munin.readthedocs.io/en/latest/? These ways people will know that the script exists and can read how to use it.

sumpfralle:
A very good point! Somehow I forgot about that.
I will add these before merging. Thanks for your thoughts!

PR #1250, opened 2019-11-21
Plugin postgres_streaming_: Use modern WAL status function names

terryburton:
PostgreSQL 10 renamed the WAL status functions as follows:

```
   - pg_current_xlog_location()      -> pg_current_wal_lsn()
   - pg_last_xlog_receive_location() -> pg_last_wal_receive_lsn()
   - pg_last_xlog_replay_location()  -> pg_last_wal_replay_lsn()
```

Use the right queries for the version of PGSQL connected to.

### DIFF
--- a/plugins/node.d/postgres_streaming_
+++ b/plugins/node.d/postgres_streaming_
@@ -96,8 +96,23 @@ my $slave  = $2;
 my $dbmaster = DBI->connect("DBI:Pg:service=$master") or die "Could not connect to master at $master\n";
 my $dbslave  = DBI->connect("DBI:Pg:service=$slave")  or die "Could not connect to slave at $slave\n";
 
-my $masterdata = $dbmaster->selectall_arrayref("SELECT pg_current_xlog_location()") or die "Could not query for xlog location on master\n";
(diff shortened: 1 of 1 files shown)


sumpfralle:
This looks good and also useful for the next 2.0 release. Thus could you please apply this change to the `stable-2.0` branch instead of `master`?

If you feel adventurous, then maybe you want to convert the `postgres_streaming_` plugin to the `Munin::Plugin::Pgsql` module (like all the other postgresql plugins)?

terryburton:
> This looks good and also useful for the next 2.0 release. Thus could you please apply this change to the `stable-2.0` branch instead of `master`?

Done: PR #1251

> If you feel adventurous, then maybe you want to convert the `postgres_streaming_` plugin to the `Munin::Plugin::Pgsql` module (like all the other postgresql plugins)?

I looked at this briefly but it doesn't look like there's a particularly good fit to the existing `Munin::Plugin::Pgsql` API, since this plugin connects to multiple databases, runs queries against each then performs a calculation returning a bespoke result. IIRC it wasn't entirely clear that the runquery subroutine was intended to be a stable part of the API rather than a private helper.

sumpfralle:
Thank you for #1251 and for considering my question regarding `Munin::Plugin::Pgsql`.

(merged via #1251)

Issue #1254, opened 2019-12-05
smart_ plugin uses same state_file for different disks configured via env.smartargs


rr-it:
smart_ plugin uses the same state_file for different disks configured via env.smartargs

**Example config to reproduce**

```
[smart_*]
user root
group disk

[smart_sg0-0]
env.smartargs -H -l error -d cciss,0

[smart_sg0-1]
env.smartargs -H -l error -d cciss,1
```
The single state-file used for all disks is `/var/lib/munin-node/plugin-state/root/smart-sg0.state`

**Expected behavior**
Disks with `env.smartargs` need the same behavior like "normal" disks: Each disk gets its own state-file as the disks are different (eg. SSD and HDD) and therefore need different state-models.

For the example config there should be two state-files:
`/var/lib/munin-node/plugin-state/root/smart-sg0-0.state` and
`/var/lib/munin-node/plugin-state/root/smart-sg0-1.state`.

**Error introduced with**
https://github.com/munin-monitoring/munin/blob/master/plugins/node.d/smart_#L501:L505

PR #1255, opened 2019-12-05
Unique state-file for each disk configured via env.smartargs

rr-it:
This fixes https://github.com/munin-monitoring/munin/issues/1254


sumpfralle:
thank you!

PR #1257, opened 2019-12-16
Plugin freeradius_sqlippool_: New plugin to show IP pool utilisation

terryburton:
Uses the generic Perl DBI interface, requiring the user to specify a driver and
provide the corresponding DBD module.


sumpfralle:
Merged via https://github.com/munin-monitoring/contrib/pull/1045 into `contrib`. Thank you!

PR #1258, opened 2019-12-16
Plugin freeradius_sqlippool_: New plugin to show IP pool utilisation

terryburton:
Uses the generic Perl DBI interface, requiring the user to specify a driver and
provide the corresponding DBD module.

### DIFF
--- a/plugins/node.d/freeradius_sqlippools_.in
+++ b/plugins/node.d/freeradius_sqlippools_.in
@@ -1,0 +1,203 @@
+#!@@PERL@@ -w
+# -*- perl -*-
+
+use strict;
+
+=head1 NAME
+
+freeradius_sqlippools_ - Plugin to monitor IP pool utilisation
+
+=head1 CONFIGURATION
+
+This is a wildcard plugin to support fetching the status of multiple sqlippool
+instances.
+
+It can also be linked directly (as with a non-wildcard plugin) to present a
+combined graph showing the percentage utilisation of all pools.
+
+It is likely that a common configuration will apply to all plugins but this
+doesn't have to be so:
+
+    [freeradius_sqlippools_*]
+        env.fr_driver mysql
+        env.fr_host 192.0.2.1
+        env.fr_port 3306
+        env.fr_db radius
+        env.fr_user radius
+        env.fr_pass radpass
+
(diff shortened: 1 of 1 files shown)


sumpfralle:
Sorry for letting your contribution rest for so long.

Your new plugin is well written and looks good to me.

At the moment there are four `freeradius`-related plugins in the munin repository itself and two other in the `contrib` repository. Thus I am not sure, which is the proper place for this new plugin. Sadly we do not have any guidelines or criteria to decide this. But my feeling is, that the topic of `freeradius` is not widespread ("common") enough to put it into the munin repository itself.

Thus I would suggest to add it to the `contrib` repository instead.

What do you think?

terryburton:
FreeRADIUS is the most popular RADIUS and AAA policy server in the world, and is also vendored in to many of the commercial products occupying that marketspace. If you look at any enterprise network (corporate wifi, 802.1X access layer, broadband access, ...) you will likely find that it is supported by FreeRADIUS or a derivative. Vague evidence of this is that you can see that the Docker image has more than 1 million downloads: https://hub.docker.com/r/freeradius/freeradius-server/

I consult for Network RADIUS who deploy this plugin in several supported installations and therefore it is very unlikely to bitrot, if that is also a concern.

Does that cause you to reconsider? If not, I'm happy for it to land in contrib.

sumpfralle:
> Does that cause you to reconsider? If not, I'm happy for it to land in contrib.

You arguments are appreciated.
The pity is, that we do not really have explicit rules, which plugins should go into the main set of munin plugins (shipped with the maintained releases) and which plugins should go to `contrib` (to be installed manually or via `munin-get`). Thus I hesitated a bit to decide here.
But I have a strong feeling, that despite the widespread usage and importance of freeradius in a specific environment, it is not a plugin commonly to be useful for the majority of munin users.
Thus I would prefer to maintain it in the `contrib` repository instead.

It is a pity, that we do not have written rules about which plugins should go where, thus we are left to subjective decision-making (which I dislike). I am sorry for that.

Thus: would you please prepare another pull request with this plugin against the `contrib` repository? Or should I try to cherry-pick your commit over there on my own?

Thank you!

terryburton:
@sumpfralle  No problem. PR created in contrib:

https://github.com/munin-monitoring/contrib/pull/1045

PR #1261, opened 2020-01-03
"munin-run" with systemd properties

sumpfralle:
The "munin-node" service process may be restricted by systemd properties.  In contrast, "munin-run" was executing the plugin with unrestricted root permissions.  This was a source of confusion for developers of munin plugins.

From now on "munin-run" tries to detect the locally configured systemd properties of the "munin-node" service ("systemctl show munin-node") and applies these to "munin-run" itself via "systemd-run".

Open questions:
* Environment variables of the `root` user are currently exposed as commandline arguments for `systemd-run`. I will fix this via another parameter `--env-file` for `munin-run` (as soon as we agree, that this is the proper way to go).
    * Update: fixed by creating a temporary `EnvironmentFile` (readable only for the caller).
* The change contains a hard-coded path (`/run/systemd/system`) in order to detect systemd. This path is used in Debian for "is systemd active?" tests.
    * Update: this path is officially specified by [systemd](https://www.freedesktop.org/software/systemd/man/sd_booted.html).
* The change contains a hard-coded systemd service name (`munin-node`). Maybe other distributions use a different name.
    * Update: the same name is used by RHEL and probably also by other distributions. In case of conflicts, distributions should patch this string or we introduce an entry in `Munin::Common::Defaults`.

@steveschnepp, @h01ger: do you have any comments or thoughts?

### DIFF
--- a/node/Build.PL
+++ b/node/Build.PL
@@ -12,7 +12,10 @@ my $build = NodeBuilder->new(
     dist_author    => 'The Munin Team <munin-users@lists.sourceforge.net>',
     dist_abstract  => 'The Munin Node',
     license        => 'gpl',
-    script_files   => '../build/node/_bin/munin-get',
+    script_files   => [
+        '../build/node/_bin/munin-get',
+        'bin/munindoc',
+    ],
     requires       => {
         perl            => '5',
         'Net::Server'   => 0,
--- a/node/sbin/munin-run
+++ b/node/sbin/munin-run
@@ -10,6 +10,7 @@ use warnings;
 # Trust PERL5LIB from environment
 use lib map { /(.*)/ } split(/:/, ($ENV{PERL5LIB} || ''));
 
+use File::Temp;
 use Getopt::Long;
 
 use Munin::Common::Defaults;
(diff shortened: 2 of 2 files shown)


kimheino:
I see that you pick only specific four hardening flags. That works, but I recommend adding following to the list:

```
Limit*  (prefix-match instead of exact)
Private*
Protect*
CapabilityBoundingSet
User
Group
UMask
```

sumpfralle:
> I see that you pick only specific four hardening flags. That works, but I recommend adding following to the list: [..]

I added the flags according to your recommendation as well as some more.

sumpfralle:
From my point of view the proposal working and complete now.

Any comments (especially regarding perl coding conventions) are appreciated.

steveschnepp:
I'd make that target `master` first, and then back port it after a while.

sumpfralle:
> I'd make that target master first, and then back port it after a while.

I doubt a bit, that it would receive enough exposure in master before the 3.0 release.
But maybe munin user are braver than I expect them to be :)

PR #1263, opened 2020-01-12
Visualize warning thresholds as dashed line

sumpfralle:
Previously warning lines used the same color, as the original.
The code also refers to a red default color in case of "single value"
graphs, but this distinction was disabled before munin 2.0 (02ff349a).

Thus currently warning threshold lines are visually hard to distinguish
from their corresponding data graph (e.g. for the "df" plugin in case of
constant space usage for read-only hosts).

This changeset introduces the formatting of warning threshold lines with
dashes.  This eases the visual distinction between data lines and
threshold lines.


sumpfralle:
# Original style (normal line)
munin_warning_non-dashed

# Proposed style (dashed line)
munin_warning_dashed

sumpfralle:
@steveschnepp: what do you think about this change for munin 2.0?

PR #1270, opened 2020-01-22
Wait for 5s before trying to connect to node

steveschnepp:


sumpfralle:
Thank you for the hint, but even a 30 seconds sleep did not help. It remains a mystery :(

PR #1282, opened 2020-02-15
munin-node: add "sd_notify" interface

sumpfralle:
The [sd_notify](https://www.freedesktop.org/software/systemd/man/sd_notify.html) interface can be used to signal the readiness of a service to its caller.

This feature can be used with `systemd` and `start-stop-daemon` in order to avoid timing issues during service startup.


sumpfralle:
@steveschnepp: would you please take a look and check whether the code is *safe* (i.e. does not break the startup under any circumstances)?

PR #1287, opened 2020-03-05
Add support for multiple fail2ban instances.

sebastic:
When multiple fail2ban instances are active on the system the path to the configuration directory needs to be supplied to the client.

This change adds the `config` variable and uses `/etc/fail2ban` by default.

### DIFF
--- a/plugins/node.d/fail2ban
+++ b/plugins/node.d/fail2ban
@@ -15,6 +15,7 @@ The following is the default configuration
 
   [fail2ban]
   env.client /usr/bin/fail2ban-client
+  env.config_dir /etc/fail2ban
 
 The user running this plugin needs read and write access to the
 fail2ban communications socket.  You will need to add this:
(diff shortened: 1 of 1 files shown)


sumpfralle:

I have two recommendations. But you may feel free to ignore both, if you prefer.

* Maybe `config_dir` would be a more suitable name for the setting.
* Maybe you want to move the fail2ban-client execution into a separate function (e.g. `run_fail2ban() { "$client" -c "$config" "$@"; }`).

I think, this change would also be useful for the `stable-2.0` branch. Thus feel free to base your changes on that branch instead, if you like. The `stable-2.0` branch is merged into `master` from time to time.

Thank you!

sebastic:
Rebased to include the two suggestions.

Retargetting to stable-2.0 will require a separate PR.

sumpfralle:
> Rebased to include the two suggestions.

Great - thanks!

A minor style nitpick: the trailing `;` in `run_fail2ban` was only required in my *everything in one line* format. But you may feel free to keep it as it is.

> Retargetting to stable-2.0 will require a separate PR.

Feel free to do so, if you would like to see this change in munin 2.0. In this case you can close this PR against `master`. You decide ...

Thanks!

sebastic:
> > Rebased to include the two suggestions.
> 
> Great - thanks!
> 
> A minor style nitpick: the trailing `;` in `run_fail2ban` was only required in my _everything in one line_ format. But you may feel free to keep it as it is.

Since you already merged the changes, I'm not going to open a new PR for that.

As I'm used to Munin plugins in Perl I didn't notice the trailing semicolon.

> > Retargetting to stable-2.0 will require a separate PR.
> 
> Feel free to do so, if you would like to see this change in munin 2.0. In this case you can close this PR against `master`. You decide ...

Since the PR for stable-2.0 was merged....

sumpfralle:
> Since you already merged the changes, I'm not going to open a new PR for that.

oh - I thought, you would include this right away in your new PR, if you wanted to do it ...

Anyway: now I removed all trailing semicolons from shell-based plugins, Thus this minor issue of taste is solved.

Thank you!

PR #1288, opened 2020-03-07
Add support for multiple fail2ban instances.

sebastic:
Retargetting of #1287 to stable-2.0 branch instead of master.


sumpfralle:
Thank you!

PR #1292, opened 2020-03-09
Proc plugin: Allow monitoring of processes with empty cmdline

kvisle:
If /proc/pid/cmdline contains nothing, the plugin would not use the counters.

cmdline should only be needed if one have specified command line arguments,
and some processes (like kswapd0 for example) do not have any contents in its
cmdline.

This patch allows kswapd0 and friends to be monitored with this plugin.


sumpfralle:

The change looks good to me.

If you like, you could rebase this commit onto `stable-2.0`. This will make the change available for the current stable release, as well. `stable-2.0` is merged into `master` from time to time. Thank you!

sumpfralle:
I just cherry-picked your commit onto `stable-2.0` (see 9b4e9349fca967cda457376239172062e5e7df8f).

PR #1298, opened 2020-03-13
Plugin cpu (linux): emit timestamp along with the values

sumpfralle:
This allows rrdtool to properly calculate the "usage by time" values for
the different categories.
Previously a slightly non-constant polling period would lead to minor
bumps in the CPU usage graphs.

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=767100

### DIFF
--- a/plugins/node.d.linux/cpu.in
+++ b/plugins/node.d.linux/cpu.in
@@ -233,6 +233,7 @@ fi
 # Note: Counters/derive need to report integer values.  Also we need
 # to avoid 10e+09 and the like %.0f should do this.
 
+timestamp=$(date +%s)
 if [ -n "$extextextinfo" ]; then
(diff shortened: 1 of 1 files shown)


sumpfralle:
@steveschnepp: could you comment on whether the above change would fix the issue described in [d#767100](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=767100)?

I do not know the details of the timestamp handling (in munin? in rrdtool?), thus I am just guessing.

sumpfralle:
Sadly the graphs of the changed `cpu` plugin still shows the same variable sum. Thus it looks like the epoch is either not properly handed over to rrd or rrd is not basing its calculation on the timestamp.

Or is this simple the wrong approach to solve the issue?

PR #1312, opened 2020-04-13
Plugin mysql_innodb: pass argument to mysql_exec

graaff:
The argument to mysql_exec should be passed to the mysql invocation
command. Without this the plugin shows errors like:

  /usr/bin/mysql: option '--execute' requires an argument
  /etc/munin/plugins/mysql_innodb: line 188: [: : integer expression expected


sumpfralle:
uh - that was me - awful :(

Thank you for spotting and fixing this!

Issue #1025, opened 2018-08-12
plugin munin_stats does not work on master


cgzones:
The plugin `munin_stats` currently does not work with the changes in master, cause munin seems to not log the time-span for an update anymore.

PR #1318, opened 2020-04-23
add start and stop log for munin-graph

d0m84:
This patch is the same as in [d5a150fdaf0e3c0eee55a5230e17743add44fef8](http://munin-monitoring.org/changeset/d5a150fdaf0e3c0eee55a5230e17743add44fef8/munin)
It adds two additional INFO log lines to the munin-graph logfile which is used by the munin-node plugin [munin_stats](https://github.com/munin-monitoring/munin/blob/stable-2.0/plugins/node.d/munin_stats.in)

**Original commit message:**
As Christoph Biedl suggested, we should at least log the start and stop of the
cron part for munin-graph if enabled. That part was remove in 2.0.12, as commit
[4372cdf24935b45f62a93b8f55e6ccacc4b6a687](http://munin-monitoring.org/changeset/4372cdf24935b45f62a93b8f55e6ccacc4b6a687/munin)

### DIFF
--- a/master/_bin/munin-graph.in
+++ b/master/_bin/munin-graph.in
@@ -82,6 +82,9 @@ if (! graph_check_cron() ) {
 	exit 0;
 }
 
+INFO "Starting munin-graph"; 
+my $graph_time = Time::HiRes::time;
+
 # BEGIN FAST-CGI LOOP:
 my $nb_request = 0;
 my $nb_request_max = 0;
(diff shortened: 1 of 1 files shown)


sumpfralle:

In this specific case I would cherry-pick (`-x`) the commit from master to `stable-2.0` in order to document the source.

Just for the clarification: does this change solve a specific problem? For example there is #1025, but I think, it refers to the `master` branch and the above commit was merged before this bug report.

d0m84:
Yes this is exclusively for the `stable-2.0` branch.

I also don't think it solves any specific open problem or #1025  
While the reason might be the same that the plugin [munin_stats](https://github.com/munin-monitoring/munin/blob/master/plugins/node.d/munin_stats) doesn't work for the `master` branch (missing log lines or log files), the solution will look completely different for this branch.

sumpfralle:
@steveschnepp: do you have an opinion regarding this commit?

sumpfralle:
I guess, the change is not problematic.

Thank you!

Issue #1201, opened 2019-06-10
Support SNMP network interfaces by name, not index


mjbnz:
**Is your feature request related to a problem? Please describe.**
I'd like for `snmp__if_` (and err) to support specifying interface name instead of index, for devices that are not switches, and can't run munin-node (Example: UniFi Security Gateway).

The USG supports pppoe, and monitoring such an interface is only possible from first boot/connection, as subsequent reconnections of the pppoe links increments the interface index. (adds/removes of VLANs to the UniFi SDN controller causes interface index changes as well)

**Describe the solution you'd like**
Allowance of `snmp__if_` to be symlinked with an interface name (or alias as specified in vyos configuration).

One implementation as below which works for me as a proof of concept, but is certainly not the cleanest implementation:
```
--- snmp__if_.orig       2017-03-04 11:44:55.000000000 +1300
+++ snmp__if_            2019-06-10 18:50:48.623653633 +1200
@@ -130,8 +130,19 @@
        exit 0;
 }

+my ($session, $error);
+
+# SNMP needed for both config and fetch.
+$session = Munin::Plugin::SNMP->session();
+
 if ($Munin::Plugin::me =~ /_if_(\d+)$/) {
     $iface = $1;
+} elsif ($Munin::Plugin::me =~ /_if_([\w.-]+)$/) {
+    my $if_alias = $session->get_by_regex("1.3.6.1.2.1.31.1.1.1.18.", "^$1");
+    if (keys(%{$if_alias}) != 1) {
+        die "Could not determine interface number from ".$Munin::Plugin::me."\n";
+    }
+    $iface = (keys %{$if_alias})[0];
 } else {
     die "Could not determine interface number from ".$Munin::Plugin::me."\n";
 }
@@ -147,10 +158,6 @@
 my $ifEntryIn64Octets  = "1.3.6.1.2.1.31.1.1.1.6.$iface";
 my $ifEntryOut64Octets = "1.3.6.1.2.1.31.1.1.1.10.$iface";

-my ($session, $error);
-
-# SNMP needed for both config and fetch.
-$session = Munin::Plugin::SNMP->session();

 if ($ARGV[0] and $ARGV[0] eq "config") {
     my ($host,undef,$version) = Munin::Plugin::SNMP->config_session();
```

An example of input data to the `get_by_regex()` call from the USG is as follows:

```
$ snmpwalk -v2c -c public <usg-ip> .1.3.6.1.2.1.31.1.1.1.18
[...]

PR #1326, opened 2020-05-24
Allow snmp iface by name

mjbnz:
As discussed in #1201 , this pull request adds support for matching network interfaces from SNMP by name, rather than index, which is useful for devices where the index changes (such as on linux based routers like the UniFi Security Gateway).

Tested working correctly for my environment. Patch includes a small fix to allow the matching of interfaces with and `@` character, such as vlans (`ethN.VID@ethN`)

### DIFF
--- a/plugins/node.d/snmp__if_
+++ b/plugins/node.d/snmp__if_
@@ -125,9 +125,22 @@ if (defined $ARGV[0] and $ARGV[0] eq "snmpconf") {
 	exit 0;
 }
 
+my ($session, $error);
+
+# SNMP needed for both config and fetch.
+$session = Munin::Plugin::SNMP->session();
+$session->translate([-nosuchinstance => 0]);
+
 if ($Munin::Plugin::me =~ /_if_(\d+)$/) {
     # Also accept leading 0 like (snmp_router_if_01 .. snmp_router_if_24)
     $iface = int $1;
+} elsif ($Munin::Plugin::me =~ /_if_([\w\@.-]+)$/) {
+    my $ifAliasContainer = "1.3.6.1.2.1.31.1.1.1.18.";
+    my $if_alias = $session->get_by_regex($ifAliasContainer, "^$1\$");
+    if (keys(%{$if_alias}) != 1) {
+	die "Could not determine interface number from ".$Munin::Plugin::me."\n";
+    }
+    $iface = (keys %{$if_alias})[0];
 } else {
     die "Could not determine interface number from ".$Munin::Plugin::me."\n";
 }
@@ -143,12 +156,6 @@ my $ifEntryOutOctets   = "1.3.6.1.2.1.2.2.1.16.$iface";
 my $ifEntryAlias       = "1.3.6.1.2.1.31.1.1.1.18.$iface";
 my $ifEntryIn64Octets  = "1.3.6.1.2.1.31.1.1.1.6.$iface";
 my $ifEntryOut64Octets = "1.3.6.1.2.1.31.1.1.1.10.$iface";
-
-my ($session, $error);
-
-# SNMP needed for both config and fetch.
-$session = Munin::Plugin::SNMP->session();
-$session->translate([-nosuchinstance => 0]);
 
 if ($ARGV[0] and $ARGV[0] eq "config") {
     my ($host,undef,$version) = Munin::Plugin::SNMP->config_session();
--- a/lib/Munin/Node/Service.pm
+++ b/lib/Munin/Node/Service.pm
@@ -50,7 +50,7 @@ sub is_a_runnable_service
     return if $file =~ m/^\./;      # Hidden files
     return if $file =~ m/\.conf$/;  # Config files
 
-    return if $file !~ m/^[-\w.:]+$/;  # Skip if any weird chars
+    return if $file !~ m/^[-\w\@.:]+$/;  # Skip if any weird chars
 
     foreach my $regex (@{$config->{ignores}}) {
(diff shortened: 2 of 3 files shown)


plugins/node.d/snmp__if_
```diff
@@ -125,9 +125,21 @@ if (defined $ARGV[0] and $ARGV[0] eq "snmpconf") {
...
+
+# SNMP needed for both config and fetch.
+$session = Munin::Plugin::SNMP->session();
+$session->translate([-nosuchinstance => 0]);
+
 if ($Munin::Plugin::me =~ /_if_(\d+)$/) {
     # Also accept leading 0 like (snmp_router_if_01 .. snmp_router_if_24)
     $iface = int $1;
+} elsif ($Munin::Plugin::me =~ /_if_([\w\@.-]+)$/) {
+    my $if_alias = $session->get_by_regex("1.3.6.1.2.1.31.1.1.1.18.", "^$1\$");
```
sumpfralle, line 138: Maybe you could assign this SNMP identifier to a variable with a descriptive name? This will make it probably easier for future readers to understand, what this line is supposed to do.

sumpfralle:

From my point of view, it looks good and useful.

Please take a look at my comment regarding the OID.

plugins/node.d/snmp__if_
mjbnz: ahh, agreed. I should have done that from the outset. Will fix.

mjbnz:
> Please take a look at my comment regarding the OID.

I've pushed a fix for this - you're right, it shouldn't be used without some context like that.

the ifAlias OID is defined later for the entry, but it has the interface index appended, so it's not really useful at this early stage when we don't know the index yet.

Therefore, I've defined it in the same scope as where it's used.. not 100% sure that's the right solution - would appreciate your input.

sumpfralle:
> Therefore, I've defined it in the same scope as where it's used.. not 100% sure that's the right solution - would appreciate your input.

It is fine for me.

In general: you may feel free to force-push your (squashed) commits (with later fixes) in your branch/pull-request. But this is just a matter of taste. It is fine, as it is now.

Now I would give others the chance to raise their voice. Otherwise I will merge your proposal in the next days.

Thank you!

mjbnz:
> In general: you may feel free to force-push your (squashed) commits (with later fixes) in your branch/pull-request. But this is just a matter of taste. It is fine, as it is now.

I'd considered it, but wasn't sure of the upstream preference. I've done that now.

And thanks!

sumpfralle:
I just cherry-picked your commits into `stable-2.0`.

Please take a look, whether my related additions to the plugin's configuration is correct from your point of view (see ded3361a4f48d0459d5a4d9f496cc4e41b80900a).

mjbnz:
> I just cherry-picked your commits into `stable-2.0`.

Awesome!
 
> Please take a look, whether my related additions to the plugin's configuration is correct from your point of view (see [ded3361](https://github.com/munin-monitoring/munin/commit/ded3361a4f48d0459d5a4d9f496cc4e41b80900a)).

Yep, that covers it off nicely. The only addition I could suggest is just adding a small comment behind the reasoning for the extra feature: `ifIndex` is not always consistent on some devices (linux based routers, basically)

sumpfralle:
> The only addition I could suggest is just adding a small comment behind the reasoning for the extra feature: ifIndex is not always consistent on some devices (linux based routers, basically)

I stared at ded3361a4f48d0459d5a4d9f496cc4e41b80900a for a minute, but failed to find a proper place to add such a comment :(

I would be happy, if you could propose such a comment as a pull request or simply write it down here in this discussion (mentioning a suitable location for such a comment). Thank you!

mjbnz:
> I stared at [ded3361](https://github.com/munin-monitoring/munin/commit/ded3361a4f48d0459d5a4d9f496cc4e41b80900a) for a minute, but failed to find a proper place to add such a comment :(
> 
> I would be happy, if you could propose such a comment as a pull request or simply write it down here in this discussion (mentioning a suitable location for such a comment). Thank you!

Yes, I wasn't terribly helpful - How about:

```
The switch port may be a number or the interface name
(e.g. "ethN.VID@ethN" - primarily useful on devices where
the interface number is not guaranteed to be consistent,
such as is the case with some Linux based systems.)
```

Definitely not required - I'm just cognizant of the fact that the question "Why was this done this way?" is often asked, and the answer has been lost to the depths of time.

sumpfralle:
Thanks!

I added this comment via cf0ff825e8a1642eede5a6bfb704790cc55fa74d.

sumpfralle:
@mjbnz: I just stumbled upon this line: https://github.com/munin-monitoring/munin/blob/fadba6315ab5d3ad4986d3ac0fa0c13175d32672/master/lib/Munin/Master/Node.pm#L348

This regular expression looks like, it should be updated according to your commit 78c7071.
Could you please check your `munin-update.log` file on your master and verify, whether you encounter lines like [this](https://github.com/munin-monitoring/munin/blob/fadba6315ab5d3ad4986d3ac0fa0c13175d32672/master/lib/Munin/Master/Node.pm#L404)?

mjbnz:
No, I don't see any `WARNING` or `ERROR` lines at all. What conditions will cause that error to be generated?

sumpfralle:
Sorry for the noise :(
I thought, the regex would be applied to the plugin name, but it is applied to `multigraph` names. Everything is fine ...

mjbnz:
Oh, no problem at all.. If you don't ask, you don't know... ;)

PR #1327, opened 2020-05-24
Allow overriding of network interface alias from SNMP

mjbnz:
A small PR to add support for overriding the interface alias from SNMP.

An example would be a plugin config file like this:

    [snmp_hostname.fqdn_if_2]
    env.alias WAN

### DIFF
--- a/plugins/node.d/snmp__if_
+++ b/plugins/node.d/snmp__if_
@@ -155,11 +155,12 @@ if ($ARGV[0] and $ARGV[0] eq "config") {
 
     print "host_name $host\n" unless $host eq 'localhost';
 
-    my $alias = $session->get_single($ifEntryAlias) ||
+    my $alias = $ENV{alias} ||
+      $session->get_single($ifEntryAlias) ||
       $session->get_single($ifEntryDescr) ||
 	"Interface $iface";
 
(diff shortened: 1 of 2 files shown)


plugins/node.d/snmp__if_
```diff
@@ -155,11 +155,12 @@ if ($ARGV[0] and $ARGV[0] eq "config") {
 
     print "host_name $host\n" unless $host eq 'localhost';
 
-    my $alias = $session->get_single($ifEntryAlias) ||
+    my $ENV{alias} ||
```
sumpfralle, line 158: Maybe you omitted `$alias = ` here?

sumpfralle:
Thank you for your proposal!

@steveschnepp: I do not really have a feeling for the SNMP plugins. Do you have an opinion regarding the suggested approach?

plugins/node.d/snmp__if_
mjbnz: Dang it, you're absolutely right - I'll fix that up.

sumpfralle:
I just cherry-picked your commits into `stable-2.0`.

Please take a look, whether my related additions to the plugin's configuration is correct from your point of view (see ded3361a4f48d0459d5a4d9f496cc4e41b80900a).

PR #1328, opened 2020-05-24
Plugin apc_nis: add UPS load in watt and hide frequency

droscy:
I've improved the apc_nis plugin disabling the print of frequency when the UPS doesn't report line frequency (like my one) and adding the UPS load in watt when the nominal power is reported by the UPS.


sumpfralle:
This looks like a very useful addition to me. Thanks!

PR #1329, opened 2020-05-27
[snmp] Update snmp__if_err_ and snmp__if_multi titles to better match…

shtrom:
… other plugins

This allows for better grouping, and more unified look, of graphs
showing the same metric collected by either SNMP and non-SNMP plugins
across different hosts (e.g., if_ and if_err_).


sumpfralle:
This looks like an improvement to me. Thank you!

PR #1330, opened 2020-06-03
Tone down lime color for graphs

pstef:
#CCFF00 is too close to color white and it's not just my opinion, see for example https://larkinweb.co.uk/computing/changing_munin_graph_colours.html


sumpfralle:
Thank you for your suggestion!

I just added a minimal script (`contrib/export-palette-to-svg`) in order to help discussing such proposals.

The two images below are the result of such an export. The first palette is the current state. The second one is the proposed set of colors.

palette-now
palette-proposed

Personally (being less sensitive for red and green), I cannot judge such a proposal.

@steveschnepp: do you have an opinion?

(or anyone else - please chime in ..)

pstef:
Here's another approach to the problem: https://martin.ankerl.com/2009/12/09/how-to-create-random-colors-programmatically/
I'm using this method in individual custom plugins with moderate success (at least I have full control over brightness).

sumpfralle:
> Here's another approach to the problem: https://martin.ankerl.com/2009/12/09/how-to-create-random-colors-programmatically/

Yes, using an evenly spread range of colors based on human perception is certainly a good approach. The perfect approach would probably something like using the CIELAB colorspace (numeric distance between color positions is equivalent to the perceived color distance) for determining well-positioned and easy distinguishable colors.

I see the following options for going forward:
* A) Allow to define acustom palette in the munin master configuration file - that should should not be too hard. This would solve the issue for you, since you would just specify whatever set of colors is suitable for you.
* B) Generate a well-balanced color palette with a "proven" (by some metric - maybe via CIELAB; or maybe some established research or widely used convention) even distribution of colors for optimal perceived distances between all colors. This palette could be added to the set of palettes distributed in munin (e.g. to be selected via `palette foo` in the munin master configuration).
* C) Same as (B), but with the goal of becoming the new default palette. Replacing the current default palette would require (IMO), that the order of colors more or less matches the current one, in order to preserve the current visual appearance as much as possible.

What do you think?

pstef:
A) is a good work-around, because I imagine it to be easy to implement and easy to use
B) I think eliminates the need for a work-around, but if I'm wrong, then A+B would be the perfect solution
C) makes it more dificult to solve a problem due to the requirement. As I understand, we would want to change the default palette so that there aren't any too bright colors, but at the same time we don't want to change anything else (most importantly, the order of the more or less matching colors).

yunal:
Hi. I made the "new" palette, I guess ~15 years ago, after corresponding with a color blind user, I forget what his weakness was. He thought the new one was better than the old. The change proposed here is totall acceptable and I'll just merge it now.  If someone comes up with a more scientific approach that is better for a lot of people that would be welcome I think.

I have seen visualization tools that offer alternative palettes on the fly for people with different kinds of color blindness.

yunal (approved):
This is good. Color is still lime and has slightly better contrast with white.

PR #1337, opened 2020-06-16
Add doc about impossibility of sharing name between multigraph sub gr…

shtrom:
…aphs and series

I just managed to debug a gnarly issue with one of my multigraphs plugins.

Adding to the doc with log messages to help troubleshooting in the future.


sumpfralle:
Thank you!

PR #1338, opened 2020-06-20
Make cpu idle time appear as blank.

22decembre:


sumpfralle:
Thank you for your proposal!

I guess, you intend to remove the metric `idle` in order to optimize the visual representation of the graph?

I have the feeling, that the plugin should not assume a numerical relationship between the given data fields (e.g. "all fields sum up to 100%"), but instead stick to the values provided by the primary source (kernel statistics). Values based on derivations or assumptions could be misleading. I think, these concerns outweigh the potential improvements regarding the visualization.

Thus I would prefer to keep collecting this data field. Or maybe I misinterpreted your goals?

22decembre:
>I guess, you intend to remove the metric idle in order to optimize the visual representation of the graph?

I think yes. More exactly, I want what happens currently with the plugin modified this way. You can have a look on my own statistic :

https://blackblock.22decembre.eu/munin/dina/dina/cpu-day.png

(If it's what you meant, then I am glad you understood me.)

I like that it clearly shows that a large part of cpu power remains unused. Except for those two picks because I launched a video encoding task that hammered it.

I would not be against keeping idle in the statistic on the assumption that the area would be filled with blank.

As of now, the way the plugin runs, I get a massive purple for idle, which I feel like it's misleading.

sumpfralle:
OK - so we are talking about the same thing :)

I could imagine, that we could introduce a configuration parameter, e.g.:
```
[cpu]
env.hide_idle_time 1
```

The plugin could omit drawing the idle time, if this environment variable is set.

If you feel like that would be a valuable improvement, then how about proposing such an addition?
In this case, I would recommend to base your commit on the `stable-2.0` branch.

I think, we will not merge the removal of this metric. But configurability is certainly fine (and looks useful).

sumpfralle:
An afterthought: if you want to be a bit more generic, then you could introduce something like `env.ignore_fields idle steal` (instead of `env.hide_idle_time 1`).

22decembre:
I think it could be ok. I mean, you're right, I prefer having a choice to do the thing I want rather than imposing my views on others. Free choice is good. Therefore I am all for configuration.

sumpfralle:
I just pushed the relevant changes in order to allow hiding specific fields (4820373).

Thank you for your inspiration!

Issue #1350, opened 2020-09-22
adjust_threshold() does not work when awk is mawk


toreanderson:
**Describe the bug**

`adjust_threshold()` from `$MUNIN_LIBDIR/plugins/plugin.sh` does not work correctly when `awk` is provided by `mawk`, as is the default case on (at least) Debian 10.

**To Reproduce**
1. Start a basic Debian 10 container: `podman run --rm -it debian:10`
2. Verify that the default `awk` is `mawk`: `awk -W version` → `mawk 1.3.3 (…)`
3. Install `munin-node`: `apt update && apt -y  --no-install-recommends install munin-node`
4. Import the `plugin.sh` library: `source /usr/share/munin/plugins/plugin.sh`
5. Test the `adjust_threshold()` function: `adjust_threshold 25% 200`

**Expected behavior**

Output from step 50 above should be `50` (i.e., 25% of 200).

**Screenshots & Logs**

Incorrect output from step 5 above:

```console
$ adjust_threshold 25% 200
4
```

**Desktop (please complete the following information):**
 - OS+Distribution Version: *Debian 10*
 - Browser: *N/A*
 - Munin Version: *2.0.49-1*

**Additional context**

Simply installing `gawk` resolves the problem (due to the fact that `gawk` has a higher priority than `mawk` in the Debian `update-alternatives(1)` system).

To reproduce the problem on a system with both `awk` implementations installed, it is possible to use a specific `awk` implementation in the following manner:

```console
$ alias awk=gawk; source /usr/share/munin/plugins/plugin.sh; adjust_threshold 25% 200
50
$ alias awk=mawk; source /usr/share/munin/plugins/plugin.sh; adjust_threshold 25% 200
4
```

kjetilho:
the problem with the code is that it uses a 0 based index for the string, but AWK is 1 based.  this means the behaviour of `substr(s, 0, n)` is undefined.
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=649147

niclan:
@kjetilho : since you taught me awk some 25 years ago: what is the correct fix?

sumpfralle:
What a nice spontaneous and productive Norse gathering :)

The issue is fixed with @kjetilho's commit (#1351).

PR #1351, opened 2020-09-22
adjust_threshold: strings indexes are 1 based in AWK

kjetilho:
GAWK accepts index 0 as an alias for index 1 in substr(), but mawk
does not.  no reason to rely on undefined behaviour.

this fixes issue #1350


niclan (approved):

sumpfralle:
What a subtle differentiation of behavior!

I merged your commit into `stable-2.0` - thus it will reach `master` soon.

Thank you!

PR #1355, opened 2020-10-21
plugins/memory: add per-cpu counter

jdh28:
Linux has a separate per CPU counter. I've observed a leak in this, which was
incorrectly being attributed to apps.


sumpfralle:
Thank you!

I merged your commit into the `stable-2.0` branch. It will reach the `master` branch soon.

PR #1359, opened 2020-11-21
Make sensors_ temperature plugin work on musl libc

tofurky:
musl libc (e.g. Alpine Linux, OpenWrt) behaves slightly differently than
glibc when LC_ALL=C is passed to the sensors command. Instead of
substituting the degree symbol for a space, it replaces it with an
asterisk. Adjust the regexes accordingly.

Also, change the autoconf regex to be consistent with the ones used to
actually parse the 'sensors' output.


sumpfralle:
Thank you!

PR #1363, opened 2020-12-08
Symlink and custom name support for diskstat_

wiebeytec:
Mainly needed for using /dev/disk/by-uuid/.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,32 @@
 -*- text -*-
 
+munin-2.0.64, 2020-07-09
+-------
+Summary
+-------
+
+Bugfix release.
+
+------------------
+Detailed Changelog
+------------------
+
+Lars Kruse (8):
+      Plugins snmp__if(_err)_: describe alias handling for configuration
+      Plugin cpu: add support for hiding selected fields
+      Plugins snmp__if*: clarify usage of interface names
+      munin-node-configure: remove "--exitnoterror" argument
+      LINESTACK: add support for undefined line width and restrict regex
+      Plugins ntp_kernel_*: handle alternative implementations (e.g. ntpsec)
+      Plugins ntp_kernel_*: force non-scientific notation for output
+      Plugin smart_: avoid ambiguous variable name
+
+Mike Beattie (3):
+      Allow '@' in plugin names
+      Support SNMP network interfaces by names instead of index
+      Allow overriding of network interface alias from SNMP
+
+
 munin-2.0.63, 2020-05-27
 -------
 Summary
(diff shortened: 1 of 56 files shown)


sumpfralle:
Thank you for your useful proposal!

Regarding the documentation:

> diskstat_iops_sde_:mnt:extrastorage

How exactly would this be interpreted? Specifically: how would you separate the path components from the name for the graph? (using only the colon sounds ambiguous to me)

If you like, you could rebase your commit onto `stable-2.0`. I think, it would be useful in the current stable version, too.
(`stable-2.0` is periodically merged into `master`)

wiebeytec:
The `:` is just translated to `/` (used existing function). The `_` are the separators. I had made a bug first (triggered when there were `_` in the custom name), but I force-pushed over it (and Github instantly redid the CI for this pull request, so perhaps you never saw it?).

This:

`diskstat_iops_sde_:mnt:extrastorage`

is matched with:

`diskstat_([-+:a-zA-Z0-9]+)_([-+:a-zA-Z0-9]+)(?:_([-+:\w]+))?$/;`

The first two matching groups are non-optional, and the last _asdf_asdf_asdf_sdf_ is captures as `asdf_asdf_asdf_sdf_`

I hadn't looked at the branches. I can rebase it there. I'll get back to it soon, I have to go now.

wiebeytec:
As Github noted, I force-pushed the branch, with the commit rebased onto stable-2.0.

It's a `.in` file there though, and the resulting file is different, because master contained some white space and spelling fixes, but the diff does apply in the same way.

sumpfralle:
Thanks for your explanation of the wildcard name scheme!
It would be great, if you could add this to the plugin documentation (I think, the usage is not obvious at the moment).
I assume, your change is backwards-compatible? (does not break current users)

Regarding the rebase: right now you rebased *all* commits from `master` to `stable-2.0`. This is a bit too much :)
A better approach would have been one of the following:
* A) `git checkout stable-2.0 && git cherry-pick YOUR_BRANCH`
* B) `git rebase --onto stable-2.0 HEAD~ HEAD`

In order to fix the current situation, I would recommend to throw away the branch, switch to `stable-2.0` and run `git cherry-pick YOUR_ORIGINAL_COMMIT`. Your original commit ID can be found in `git reflog`.
Sorry for the mess ...

wiebeytec:
The change is backwards compatible indeed.

I think what happened with git, is that the force push re-linked the pull request to master, but that's not valid, because it's supposed to go to stable-2.0 now. My own git clone looks proper. I'll update the documentation as requested, and will make a new pull request from a new branch name. That means I'll close this, but I'll be back :)

sumpfralle:
Great!

PR #1366, opened 2020-12-16
fix reporting on URL in the munin-update log

Sephtex:
This will fix the reporting of the URL in munin-update.log in function of the protocol used.

Previously ssh and cmd reported a port at the end in the logs which was confusing. This will report the address how we expect it to be used.

### DIFF
--- a/lib/Munin/Master/UpdateWorker.pm
+++ b/lib/Munin/Master/UpdateWorker.pm
@@ -65,20 +65,20 @@ sub do_work {
     my $path = $self->{host}->get_full_path;
     $path =~ s{[:;]}{-}g;
 
-		# Parameters are space-separated from the main address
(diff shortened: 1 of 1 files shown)


sumpfralle:
This looks good!
How about applying this to `stable-2.0` instead? (just cherry-pick and force-push)

`stable-2.0` is periodically merged into master.

Thank you!

Sephtex:
Sure, will do it imediately, sicne stable-2.0 is periodically merged into master, I suppose we can decline this PR?

Sephtex:
https://github.com/munin-monitoring/munin/pull/1368 was created from stable-2.0 to stable-2.0

sumpfralle:
Closing, since #1368 is merged now, thanks!

PR #1368, opened 2020-12-16
fix reporting on URL in the munin-update log (stable-2.0)

Sephtex:
This will fix the reporting of the URL in munin-update.log in function of the protocol used.

Previously ssh and cmd reported a port at the end in the logs which was confusing. This will report the address how we expect it to be used.

Replacing PR of https://github.com/munin-monitoring/munin/pull/1366 to master


sumpfralle:
Thanks!

PR #1374, opened 2021-01-14
Make netstat human and high connection friendly

wiebeytec:
Changes Y-axis to use SI units instead of 5e-02 till 1e+02 and such.

Makes it multigraph with a second linear graph for 'established',
because that number can be very much higher and natural variations will
just be a flat line in log scale.

Remove the arbitrary limit of 50000 max established connections.

Before:

Munin-original-netstat

After example 1:

MuninNetstatFixed1

After example 2:

MuninNetstatFixed2

### DIFF
--- a/plugins/node.d.linux/netstat.in
+++ b/plugins/node.d.linux/netstat.in
@@ -65,8 +65,9 @@ fi
 
 if [ "$1" = "config" ]; then
 
-	echo 'graph_title Netstat'
-	echo 'graph_args --base 1000 --logarithmic'
+	echo 'multigraph netstat'
+	echo 'graph_title Netstat, combined'
+	echo 'graph_args --units=si -l 1 --base 1000 --logarithmic'
 	echo 'graph_vlabel TCP connections'
 	echo 'graph_category network'
 	echo 'graph_period second'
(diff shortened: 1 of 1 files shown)


wiebeytec:
I just realized something I haven't tested yet. Because I developed it as separate plugin, I don't know what it does with history when it replaces the old netstat plugin. Testing it now, but from the look of the data dir with rrds, it will loose all history :(

wiebeytec:
> I suppose by calling the first multigraph 'netstat', it will take the history?

Confirmed. Altered commit, force pushed. The pull request has been updated automatically.

niclan:
If you go "multigraph netstat" and then use the same labels for that graph as the traditional plugin (and remove the traditional one) it should take over the old ones rrds and continue on them.

wiebeytec:
> If you go "multigraph netstat" and then use the same labels for that graph as the traditional plugin (and remove the traditional one) it should take over the old ones rrds and continue on them.

Yep, I managed it. Sorry for the comment spam to myself:
Munin-with-netstat-history

Never mind the graph title; that's not part of the pull request.

sumpfralle:
Yes, that graph looks much better now!
Thank you for your contributing and for taking the time to fix the backwards compatibility!

PR #1394, opened 2021-04-05
apt_all: remove reference to "update" argument for cron job

sumpfralle:
Before commit 904a84e392ed2 the plugin allowed to be called with an
"update" argument in order to trigger a package cache update.

The plugin does not support this anymore.
Instead the infrastructure provided by the system ("APT::Periodic")
should be used instead.
This is now clarified in the plugin documentation.


sokai:
FYI: _#fixed_ (dup → #1420)

PR #1397, opened 2021-04-21
Use the host-specific local_address configuration if present

wferi:


sumpfralle:
Good catch!

I merged this (manually, de74370848a7d18ce780dfcf8dbd8dbb026f6128) into `stable-2.0` (instead of `master`) now, since it is already documented to work.

Thank you!

wferi:
Great, thanks!

PR #1399, opened 2021-04-30
use dedicated munin SNMP plugin

guillomovitch:
This reduce code duplication, and honour configuration variables, such
as SNMP version to use.


sumpfralle:
Thanks for cleaning this up!

PR #1401, opened 2021-05-20
threads: Use short xargs arguments for busybox compatibility

brknkfr:
The xargs command included in busybox only knows short args. (I recently installed munin on alpine linux, which is based on busybox)

### DIFF
--- a/plugins/node.d.linux/threads
+++ b/plugins/node.d.linux/threads
@@ -49,7 +49,7 @@ fi
 # read.  It isn't entirely portable, but GNU grep should be a given on Linux.
 # Sadly awk has no such equivalent option or we could skip grep altogether.
 find /proc/ -mindepth 2 -maxdepth 2 -type f -name status -print0 2>/dev/null \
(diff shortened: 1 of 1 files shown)


sumpfralle:
Interesting!

I just tried this here:
```
$ echo foo | busybox xargs --no-run-if-empty
foo
```

Thus it seems to work on Debian.

I can only guess, that the maintainer of the alpine package did not enable one of the many busybox compile flag?

I would really like to keep the long versions of these options in order to increase readability.
Maybe you could ask the alpine package maintainer about this?

brknkfr:
> Thus it seems to work on Debian.
> 
Okey, I didn't try it on other distributions.

> I can only guess, that the maintainer of the alpine package did not enable one of the many busybox compile flags?
> 
> I would really like to keep the long versions of these options in order to increase readability.
> Maybe you could ask the alpine package maintainer about this?
>
Probably. Let me check.

sumpfralle:
I am closing this suggestion for now, since I would prefer the solution in the alpine repository.

Feel free to re-open the PR, if you think otherwise. Thanks!

PR #1403, opened 2021-05-26
Fix plugins_without_multigraph_check in Makefile

maartenvs:
Fixes a typo in the `lint-plugins` target of the Makefile that caused the test for plugins without multigraph check to always pass without errors.

Tested by temporarily removing the line

    need_multigraph();

from `plugins/node.d/proxy_plugin` and running `make lint-plugins`. That should report an error.


sumpfralle:
Good catch - thanks for spotting this!

maartenvs:
> Good catch - thanks for spotting this!

Thanks! What is the next step to get the pull request merged? Is there anything I can do to help?

sumpfralle:
> What is the next step to get the pull request merged?

@steveschnepp will merge it, I guess.
(I maintain only the `stable-2.0` branch)

PR #1417, opened 2021-08-09
plugins/postfix_mailqueue: don't run postconf if spooldir-override is defined

kimheino:
spooldir-env takes priority so there's no reason to run postconf if
it's defined. It will trigger SElinux denials on some systems.


sumpfralle:
Thank you!

PR #1425, opened 2021-09-21
plugins/postgres_: enable configuring warning/critical

ap-wtioit:
for postgres_connections_ALL this enables configuring warning / critical
 values that reflect the postgres configuration on the server. to find
 suitable values one can execute `psql -c 'SHOW max_connections'`
 
 

Info @wt-io-it

### DIFF
--- a/lib/Munin/Plugin/Pgsql.pm
+++ b/lib/Munin/Plugin/Pgsql.pm
@@ -51,12 +51,18 @@ variable, but this is usually a bad idea.
 If you are using plugin for several postgres instances, you can customize
 graph title with the environment variable PGLABEL.
 
+Warning and critical can be customized by setting "warning" and "critical" env variables per plugin.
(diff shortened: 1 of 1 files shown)


ap-wtioit:
yes with this patch you can override any warning and critical settings in env for every plugin that uses Pqsql.pm. Currently it seems none of the postgres_* , pgbouncer_* or slony_lag_ plugins seem to set a warning. Overriding the warning and criticals this way seems to be used by quite a few plugins, so i think it's the right order.

We ran out of postgres connections last week so i tried to set warning and discovered that for the postgres plugins it's not configurable (i tried to figure out how to compute it dynamically but i couldn't figure that one out).

I'm, running a version of this patch with 2.0.37-1ubuntu0.1 so yes, it should be suitable for merging there. I will open a second merge request.

Issue #1424, opened 2021-09-17
iostat plugin doesn't handle NVME drives


svenedge:
**Describe the bug**
NVME drives are excluded unless SHOW_NUMBERED is enabled, because the drives as well as partitions end in digits.
```
lsblk
NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
nvme0n1     259:1    0  150G  0 disk
└─nvme0n1p1 259:2    0  150G  0 part /
nvme1n1     259:0    0   20G  0 disk
nvme2n1     259:3    0   20G  0 disk
```
The plugin already has an exception for CCISS drives, perhaps something similar could be added for NVME?
https://github.com/munin-monitoring/munin/blob/bc6450d8ebb1294fc7001f39c2cee28322c743c3/plugins/node.d.linux/iostat#L173-L177

**To Reproduce**
1. Have NVME drives
2. Run `/usr/share/munin/plugins/iostat` which won't show them
3. Run `env SHOW_NUMBERED=1 /usr/share/munin/plugins/iostat` which will, but also the partition(s).

**Expected behavior**
The NVME drives are detected by default, but not the partitions.

**Desktop (please complete the following information):**
 - OS+Distribution Version: Centos 7
 - Munin Version: munin-node-2.0.66-1.el7.noarch

sumpfralle:
> The plugin already has an exception for CCISS drives, perhaps something similar could be added for NVME?

That sounds reasonable!

Could you please provide a (dirty?) patch for this?

PR #1427, opened 2021-09-23
Handle NVME drives in the Linux iostat plugin

svenedge:
Fix NVME drives not being detected by iostat plugin, because the device names end in digits. See #1424.


sumpfralle:
Thank you for fixing this issue!
Could you please apply your commit to the `stable-2.0` branch instead? It would be beneficial for the 2.0 releases ...
Thank you!

PR #1431, opened 2021-10-15
Handle NVME drives in the Linux iostat plugin (2.0 branch)

svenedge:
Fix NVME drives not being detected by iostat plugin, because the device names end in digits. See #1424 and #1427.


sumpfralle:
Thank you!

PR #1436, opened 2021-11-13
Use `grep -F` instead of `fgrep`

scop:
`fgrep` has been deprecated in GNU grep since 2007, and in current post
3.7 Git it has been made to emit obsolescence warnings:
https://git.savannah.gnu.org/cgit/grep.git/commit/?id=a9515624709865d480e3142fd959bccd1c9372d1

### DIFF
--- a/Checklist
+++ b/Checklist
@@ -13,12 +13,12 @@ This is a pre-release checklist for the maintainer.
 
    2. Figure out which r the previous tag was at and do this (r910 was
 	  1.3.2 and r1234 is head)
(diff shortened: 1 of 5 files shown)


sumpfralle:
Thank you for spotting this!

Could you please apply your changes to `stable-2.0` instead `master`?
(it will get merged into `master` later)

scop:
Sure, done.

sumpfralle:
Cool - thank you!

PR #1443, opened 2021-12-02
UpdateWorker: prevent use of uninitialized value $first_epoch

ap-wtioit:
if {when} was a zero length array munin-update.log would show:
[PERL WARNING] Use of uninitialized value $first_epoch in subtraction
(-) at /usr/share/perl5/Munin/Master/UpdateWorker.pm line

added a check if array is not empty otherwise we use the default 0 value

This fixed https://github.com/munin-monitoring/contrib/issues/1177 on our server running Debian Buster

Info @wt-io-it


sumpfralle:
Thank you for researching and fixing this issue!

PR #1444, opened 2021-12-03
Fix link for rrdgraph graph_args

ap-wtioit:
rrdgraph_graph.en.html holds the information regarding lines and area drawing.
rrdgraph.en.html holds the information relevant to munin graph_args config option.

Info @wt-io-it


sumpfralle:
The change looks good to me.
@steveschnepp: would you merge it?

PR #1447, opened 2022-01-07
Bypassing failing SHOW PLUGINS command

muhgatus:
If your mysql version has no SHOW PLUGINS, skip this.


sumpfralle:
Thank you for your suggestion!

Could you please add a comment above the `eval`, which explains its necessity?

And could you please apply your patch on top of `stable-2.0` instead of `master`? It should be helpful for the 2.0 releases, I guess.

PR #1456, opened 2022-03-10
Exclude msdosfs as there is no inode on that filesystem

daftaupe:
It makes no sense to try to report the inode usage for a msdosfs filesystem, so I've just added this type to the list of excluded ones.


sumpfralle:
Thanks!

PR #1462, opened 2022-04-01
plugins: Add Linux PSI plugin

cdown:
PSI is the most modern and comprehensive way to measure resource
contention on Linux servers, see
https://www.kernel.org/doc/html/latest/accounting/psi.html. It's a
relatively mature kernel interface and has already proved itself to be
very useful finding and debugging issues at scale, and it makes sense to
have in the munin main repository.

Metrics present as percent of time spent doing "wasted" work due to
contention, like having to refault pages from disk. This avoids having
to guess when a workload is approaching the limit based on things like
the used memory and memory composition, which can be highly misleading.

### DIFF
--- a/plugins/node.d.linux/psi
+++ b/plugins/node.d.linux/psi
@@ -1,0 +1,92 @@
+#!/bin/sh
+
+: << '=cut'
+
+=head1 NAME
+
+psi - Plugin to graph Linux pressure stall information
+
+=head1 APPLICABLE SYSTEMS
+
+Linux kernels since 4.20 with CONFIG_PSI=y.
+
+=head1 INTERPRETATION
+
+When resources are contended, workloads experience latency spikes, throughput
+losses, and run the risk of OOM kills.
+
+Linux's PSI feature, introduced in kernel 4.20, identifies and quantifies the
+disruptions caused by resource contention.
+
+=head1 AUTHOR
+
+Copyright (C) 2022 Chris Down <chris@chrisdown.name>
+
+=head1 LICENSE
+
+GNU GPLv2
+
+=head1 MAGIC MARKERS
+
+  #%# family=auto
+  #%# capabilities=autoconf
+
+=cut
+
+psi_dir=/proc/pressure
+
+autoconf() {
+    if [ -d "$psi_dir" ]; then
+        echo yes
+    else
+        echo "no ($psi_dir not found)"
+    fi
+}
+
+config() {
+    echo "graph_title PSI"
+    echo "graph_args --upper-limit 100 -l 0"
+    echo "graph_vlabel %"
+    echo "graph_category system"
+    echo "graph_info This graph shows resource contention on the machine." \
+         "Values are recorded from the 'avg10' counter." \
+         "%full cpu is excluded as it's meaningless at the system level."
+
+    for res in memory cpu io; do
+        while read -r psi_type _; do
+            if [ "$res" = cpu ] && [ "$psi_type" = full ]; then
+                # See comment in get_data
+                continue
+            fi
+            printf '%s%spressure.label %%%s %s pressure\n' \
+                "$res" "$psi_type" "$psi_type" "$res"
+            printf '%s%spressure.info Percent %s time in "%s" contention\n' \
+                "$res" "$psi_type" "$res" "$psi_type"
+        done < "$psi_dir/$res"
+    done
+}
+
+get_data() {
+    # Example line: some avg10=0.00 avg60=0.00 avg300=0.17 total=91233216
+    #
+    # CPU has "full" to keep parsers the same as for cgroup-local PSI stats,
+    # but it's meaningless at the system level, so it's ignored.
+    for res in memory cpu io; do
+        while read -r psi_type pct_ten _; do
+            if [ "$res" != cpu ] || [ "$psi_type" != full ]; then
+                pct_ten=${pct_ten#*=}
+                printf '%s%spressure.value %s\n' "$res" "$psi_type" "$pct_ten"
+            fi
+        done < "$psi_dir/$res"
+    done
+}
+
+case ${1:-fetch} in
+    config) config ;;
+    autoconf) autoconf ;;
+    fetch) get_data ;;
+    *)
+        printf 'Unknown option: "%s"\n' "$1"
+        exit 1
+    ;;
+esac


cdown:
If I understood from reading other plugin additions to the main repo correctly, new contributors require a sponsor that already has contributions, so soliciting @Flameeyes who knows me in person.

Flameeyes:
I'm (kind of obviously, given I've not touched this in a long time) not authorized to approve the workflow, but I can definitely vouch for @cdown.

sumpfralle:
Thank you for contributing this plugin.

Just a few moments ago I merged a plugin with a similar scope into the `contrib` repository: https://github.com/munin-monitoring/contrib/blob/master/plugins/system/linux_psi

Maybe you want to take a look at it and compare both feature sets? Maybe one of these could benefit from the other one? Or maybe there are good reasons for having both separate plugins.

Regarding the inclusion of the plugin into the munin core repository (instead of `contrib`): at least I am using Linux, thus the maintenance of the plugin should not be an issue. And indeed I have the feeling, that this metric could be quite useful for almost every munin user out there. Thus I am slightly leaning towards including such a plugin (either this implementation or the other one from `contrib`).

@steveschnepp: what do you think about including a plugin for the Linux pressure stall information?

plugins/node.d.linux/psi
```diff
@@ -0,0 +1,94 @@
...
+        while read -r psi_type pct_ten _; do
+            if [ "$res" != cpu ] || [ "$psi_type" != full ]; then
+                pct_ten=${pct_ten#*=}
+                printf '%s%spressure.value %s\n' "$res" "$psi_type" "$pct_ten"
+            fi
+        done < "$psi_dir/$res"
+    done
+}
+
+case ${1-data} in
```
sumpfralle, line 86: Probably you meant `1:-` here (you missed the colon).
And please use `fetch` instead of `data` as the default (some munin implementations use this word).

steveschnepp:
> If I understood from reading other plugin additions to the main repo correctly, new contributors require a sponsor that already has contributions, so soliciting @Flameeyes who knows me in person.

yes, this is because I don't want to add more unmaintained plugins in core. 

that said, if @sumpfralle is happy to maintain it, I don't have a issue.

May I just suggest to have it in contrib for a while first ?

plugins/node.d.linux/psi
```diff
@@ -0,0 +1,94 @@
...
+        while read -r psi_type _; do
+            if [ "$res" = cpu ] && [ "$psi_type" = full ]; then
+                # See comment in get_data
+                continue
+            fi
+            printf '%s%spressure.label %%%s %s pressure\n' \
+                "$res" "$psi_type" "$psi_type" "$res"
+            printf '%s%spressure.info Percent %s time in "%s" contention\n' \
+                "$res" "$psi_type" "$res" "$psi_type"
+            printf '%s%spressure.type DERIVE\n' "$res" "$psi_type"
```
HaseHarald, line 65: As the author of "the other psi plugin", I've got to say, I'm pretty impressed by the code efficiency of this plugin. There's still something to learn for me.

I'm curious though, why you chose to use `DERIVE` for the "average" metrics. Maybe I got something wrong from the documentation (which is not unlikely), but as far as I know, these values are in percent. Is there a major benefit in watching just the rate of change? I can see that for the "total" metrics, which are only ever increasing and reach astronomical values that hide all details when viewed as is. But for the "averages"?

Also, is there a reason why you focus on the 10 second average? After all, by default munin collects metrics every 5 minutes (which can be changed of course). Therefore the 300 second average would seam more logical to me. However, I don't have a lot of experience with the psi values in general yet. So I'm probably missing something important here (which is why I'm asking).

plugins/node.d.linux/psi
cdown: `-` seems more logical than `:-`, since if someone passes `plugin ''` that should be an error. Happy to change the verb :-)

plugins/node.d.linux/psi
cdown: I chose `DERIVE` because, as I understand it, derive + min 0 prevents dropping below zero due to sampling shenanigans.

The code focuses on the 10 second average because while the counters are aggregated at 5 minutes, they are collected more frequently, as I understand it. As such, focusing on avg300 is likely to obscure resource contention spikes.

I am by no means a Munin expert though, so I may be wrong about the derive + min or collection interval stuff. :-)

cdown:
> yes, this is because I don't want to add more unmaintained plugins in core.

I am one of the contributors to cgroups and PSI in-kernel, so I'm more than happy to maintain it. :-)

> May I just suggest to have it in contrib for a while first ?

I mean, that's fine, but it would be good to know how long "a while" is and what would the preconditions would be for getting it merged into upstream. PSI is Linux's best way to represent resource pressure -- far more so than the metrics already present in core -- and not having it in mainline seems like a miss.

We've already stabilised the PSI interface upstream so there won't be changes to the format or meaning of these files.

plugins/node.d.linux/psi
HaseHarald: Well, as far as I know, derive does what it says. It derives the function. So instead of the actual value, you get the rate of change compared to the previous value(s). This is very useful for metrics, where you don't care as much about the actual value but more about it's direction.

For example the "total" metrics for the psi, which are (as far as I know) counter values. So the only ever go up. Using a derive, one can easily see when and by what rate they rise. However, once you reboot the machine, the counter gets reset and starts at zero again. For a graph with "derive" this would mean a huge negative spike, that actually has no meaning, because it's just caused by a counter reset. Therefore using `min 0` prevents this.

However when looking at the average metrics, we are looking at the percentages of an average over time. These values naturally go back to zero every now and then. Well, I guess that's an understatement and would probably be a sign of problems. Let me rephrase that: "These values are not zero every now and then". Whatever, the point is, that it's normal for these values to go back to zero and therefor would cut of that portion of the derived graph (when using `min 0`) and give a false impression. Also, since these values are in percent already, I don't see the point in using `DERIVE` at all. I personally can grasp values in percent very well. Therefore I was wondering what the benefit of deriving these metrics is. Since you are the one with way more experience about psi than I am, I figured there might be a good reason for that. So I asked.

Same goes for the choice to focus on the 10 second average. Yes the 300 second average is an average over a bigger time span and therefore looses details. On the other hand the 10 second average only shows the spikes that are present in the average in the 10 seconds before the plugin was invoked. The other 290 seconds between two plugin calls are completely ignored. I guess both have their benefits, maybe not. I'm not an expert on psi yet. For my plugin is solved this problem by including all metrics by default and giving the option to exclude ones that you don't care about by configuration. Which to be fair is a pretty lame approach, since it spared me the time for looking deeper into how psi works.

So don't get me wrong. It's not meant as bad critique. I just stumbled across these things that seamed odd to me. But since I have to admit, that I'm not an expert on the psi metrics, I figured there is probably something I'm missing and took the opportunity to learn something.

plugins/node.d.linux/psi
sumpfralle: I appreciate your desire for distinguishing between an empty argument and an unset argument. :)

But I have a few issues with this specific implementation:
* A) it *looks* like a malformed `${foo:-default}` (thus I feel the urge to fix it)
* B) I could not find a documentation in `man dash` or `man bash` for this expression (but I tried it and it works - to my surprise). Do you have a reference?
* C) `data` looks like a magic *munin* word - but it is just a magic *plugin* word
* D) The other munin plugins are rather relaxed with regard to the cleanliness of their input. I would not try to introduce a stricter behavior solely in this plugin.

Thus I would recommend `${1:-fetch}` :)

cdown: `-` is POSIX, see the table here: https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html#tag_02_06_02

If you prefer `:-` to more closely match existing plugins, that's fine. I've just sent a new version with `:-fetch`. :-)

sumpfralle: Hehe, thanks for sharing this piece of obscure knowledge! I always missed that before ...

Indeed: your usage of the hyphen is in this situation more correct, than my suggestion.
But I really cannot remember, having ever stumbled upon this type of parameter expansion anywhere (and I think, I have read a *lot* of shell code). Munin's `contrib` repository contains exactly two shell plugins (`hp2600_count_` and `hp2600_status_`), which use this parameter expansion.

Thus I think, we should prefer the (imperfect), but less exotic approach with `:-`. Thank you for changing this!

plugins/node.d.linux/psi
sumpfralle: Thank you for your exchange of thoughts!

Regarding the `avg300`: I think, this should be the proper choice here. All values will be lower, but this does not prevent us from seeing relative changes in proper proportion. The coverage of munin's default period of 300 seconds is preferable.
Could you please change this?

sumpfralle:
> May I just suggest to have it in contrib for a while first ?

@steveschnepp: I think, it would not significantly improve the test exposure of the plugin, if we would move it to `contrib` first (since there are probably only a few people watching that space). And I would prefer to *not* remove it from `contrib` just a bit later again (and cause confusion for users).

Thus I would suggest, that we call for testing in the IRC and via the user's mailinglist. This should give us enough feedback in a short amount of time.

@cdown: what do you think about adding the `multigraph` feature to this plugin in order to separate the domains `io`, `memory` and `cpu`?

cdown:
Hmm, I may not totally understand how `multigraph` works, but doesn't it mean that they are then presented separately?

If that's the case it seems non ideal, since one of the primary things about PSI and resource movements in general is that they should be looked at together. I go into this somewhat at my talk at SRECon a few years ago: https://www.youtube.com/watch?v=beefUhRH5lU&t=395s

For example, memory pressure will, eventually, turn into IO (major fault) and CPU (reclaim difficulty) pressure. Having these on the same graph makes it much easier to see what the causality is, whereas changing it to be in different graphs seems to make that harder.

Or maybe I misunderstood how `multigraph` works? :-)

sumpfralle:
Yes, you understand it properly.

I assumed, the different categories of symptoms of load could be better visualized by separating them. If you think, a single graph is preferable, then it is obviously fine as it is.

sumpfralle:
Do you think, the plugin is in a more-or-less finished state now?

Then I would ask for some testing on the mailing list and via IRC, before merging it into `stable-2.0`.

cdown:
Thanks! Yes, I don't think there is anything pending. Would you like me to send the mail to the mailing list, or would you like ti send it? :-)

sumpfralle:
> Would you like me to send the mail to the mailing list, or would you like to send it?

Thanks for the offer, but I would prefer to do it :)

After looking over the plugin again, I see three potential improvements and I would be interested in your opinion:
* changing the field names from `memorysomepressure` to `memory_some_pressure`
* use the 300 seconds average (instead of 10 seconds)
* rename the plugin:
    * currently it *sounds* like a physical (meteorological) property
    * the "other" plugin is named `linux_psi`
    * I could imagine a name like `linux_pressure` or `pressure_stall_information`
 * maybe add `local` specifications for the function variables?

What do you think?

HaseHarald:
Hello @cdown,

I'm not constantly on GitHub (partly because GitHub starts to annoy me a lot in the recent time with its changes, but that's a different topic), so I'm not always answering in time. Also I'm not actually involved in this project and take part here mainly because of interest. So, sorry it it takes me a while to answer sometimes.

Still I hope I can shed some light.

> Hmm, I may not totally understand how `multigraph` works, but doesn't it mean that they are then presented separately?
> 
> If that's the case it seems non ideal, since one of the primary things about PSI and resource movements in general is that they should be looked at together to understand causality and interactions. I go into this somewhat at my talk at SRECon a few years ago: https://www.youtube.com/watch?v=beefUhRH5lU&t=395s
> 
> For example, memory pressure will, eventually, turn into IO (major fault) and CPU (page scan) pressure. Having these on the same graph makes it much easier to see what the causality is, whereas changing it to be in different graphs seems to make that harder.
> 
> Or maybe I misunderstood how `multigraph` works? :-)

First of all, thanks for the interesting insight into your work-area. That really gave me some understanding.

About `multigraph`. It can be used to divide up graphs, yes. But you can also use it to add a drill-down view, where you have a combined view on the main dashboard and get more details once you click on the graph. For example my implementation of the psi-plugin gives you the 300 second averages (well, that's actually configurable) of all resources (cpu, memory, io) in one graph. When you click on that one, you'll get 6 more graphs, two for each resource, one for all averages and one for the rate of change (derive) for the totals. For the totals I had a single graph in mind at first. However I found that the differences in value are so huge, that a lot of detail will get lost, when combining them into one graph. Although one could argue that logarithmic scaling could help with that.

Maybe give that plugin a try and see for yourself how `multigraph` affects that. Find it here: https://github.com/munin-monitoring/contrib/blob/master/plugins/system/linux_psi

> ...
> 
> I'm curious though, why change from 10 second counters to the 300 second counters? It's more likely that spikes in resource pressure will be missed that way, or the exact time that a resource pressure event started will beomc less clear, so we usually recommend using the shorter ones.
> 
> ...

The idea of using the 300 second average probably stems from the fact that that's the default sample rate of Munin. Maybe that's the source of misunderstanding in the first place? To clarify on that: Munin is not a monitoring tool with constant data collection. Instead it relies on polling (like most other monitoring tools as well, I might add), and the default polling rate is 5 minutes. So choosing the 10 second average might improve the spike detection in some cases, but will have you missing a lot of spikes in general.

Let's do an example: Say it's 01:00am and Munin does it's polling. It grabs the 10 second average value for all resources and plots them to it's graph. They are pretty low, no need to get alarmed, because the machine was not actually doing anything in the last 10 seconds. The next polling run of Munin is at 01:05am and again it looks for the 10 second averages of all resources. Again these values are pretty low and it plot's that to it's graph. When you look at your graph, you'll see a nice flat line, nothing to worry about. What you missed is the huge spike that lasted from 01:01:12am to 01:02:53am, because Munin wasn't actually polling in that time and you are only asking for the 10 second average. So you are only ever looking at the values in the time frame 10 seconds before the Munin poll. In the 300 second averages, these values would be included though. They don't give such a huge spike in the graph and you don't get an exact time for the spike, but I guess it's better than missing it completely.

Hope that answers some questions.

Kind Regards

Hase Harald

sumpfralle:
@cdown: thanks for your patience! I will happily wait for your changes.

@HaseHarald: thank you for jumping in and explaining munin's approach here!

Indeed the `multigraph` feature of munin allows two different approaches:
* A) emit multiple (independent) graphs from a single plugin
    * this is, what I tried to suggest with `memory`, `io` and `cpu` before
* B) emit a nested set of graphs (e.g. for providing more details at the lower level)
    * this is @HaseHarald's suggestion (e.g. for different *averaging periods*)

I think, the two of you have a better understanding of the PSI feature than I, thus I would trust you to choose a helpful set of graphs here.
And of course, the graphs and fields of this plugin will not be set in stone forever - but I would like to start with a good initial set :)

Maybe @HaseHarald's explanation of munin's polling periods and his suggestion for possible nested graphs inspire some thoughts in @cdown? I am looking forward to your exchange.

Thank you for you thoughts and time!

cdown:
Sorry for the delay folks, the changes are finished, but then I wanted to look at this `multigraph` business since it does sound useful and didn't have time to test it properly yet with other upstream changes. I'll send again with `multigraph` soon. :)

PR #1470, opened 2022-05-17
Don't suggest ipmi_ links if ipmitool isn't runnable on this system

spacelama:
Supress unconditional suggestion of _fans _temp _power _volts if ipmitool isn't runnable

### DIFF
--- a/plugins/node.d/ipmi_
+++ b/plugins/node.d/ipmi_
@@ -43,18 +43,25 @@ Donated to the public domain by Nicolai Langfeldt (janl@linpro.no)
 
 CONFIG=no
 
+autoconf() {
+    type -p ipmitool &>/dev/null ||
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for taking a look at this plugin!

I think, it is not necessary to guard `suggest` with a check for the result of `autoconf`, since this is supposed to be done by the caller of `suggest` (see the function `fetch_plugin_suggestions` in `node/sbin/munin-node-configure` or the [description of `suggest`](http://guide.munin-monitoring.org/en/latest/develop/plugins/plugin-concise.html#suggest)). Thus the `suggest` implementation of a plugin may safely assume, that `autoconf` results in a success.

Or what do you think?

btw.: changes like these should go into the `stable-2.0` branch, I guess. Sadly this is not obvious. Just as a hint for your future contributions ...

Thanks!

PR #1471, opened 2022-05-19
Plugin bind9: stabilize graph + minor change

penguinpee:
Stabilize graph of bind9 plugin applying same fix as for
bind9_rndc plugin in 2c72c916.

Update URL of author's blog.

Change graph type and prioritize important data.

### DIFF
--- a/plugins/node.d/bind9.in
+++ b/plugins/node.d/bind9.in
@@ -92,7 +92,7 @@ sub do_stats {
     get_state;
 
     open(Q,"> $STATEFILE") or die;
-    foreach $k (keys %IN) {
+    foreach $k (sort keys %IN) {
         print "query_$k.value ",$IN{$k},"\n";
         print Q "$k ",$IN{$k},"\n";
     }
@@ -115,7 +115,7 @@ query_other.draw AREA
 ";
     get_state;
 
-    foreach $k (keys %IN) {
+    foreach $k (sort keys %IN) {
         print "query_$k.label $k
 query_$k.type DERIVE
 query_$k.min 0
--- a/plugins/node.d/bind9.in
+++ b/plugins/node.d/bind9.in
@@ -108,10 +108,6 @@ sub do_config {
     print "graph_title DNS Queries by type
 graph_category dns
 graph_vlabel Queries / \${graph_period}
-query_other.label Other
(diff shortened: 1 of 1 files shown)


steveschnepp:
Looks good to me. 

The only concern I have is that the order will change for order. But that is moot as others are random anyway.

So, all good.

penguinpee:
> The only concern I have is that the order will change for "other".

I don't quite follow (early morning still). "Other" will move to the bottom. The known values will be ordered alphabetically. Before they were random and "Other" always at the top, presumably due to the use of a stacked graph. I can provide before and after pictures of the graph, if that helps.

steveschnepp:
ah. usually random things should be at the top. Otherwise they have the tendency to "pollute"

sumpfralle:

(I am following @steveschnepp's preference regarding "other", but your contribution is a very welcome improvement, just as it is)

penguinpee:
Glad I could help.

I'm still trying to wrap my head around, what you mean by *random* in connection with "other" category. It's not random. None of the values are. Other may be unpredictable, but it's not random. It's a catch all for everything that's not an internet (IN) query or otherwise malformed, for whatever reason.

But, thanks for including this.

steveschnepp:
Usually, if you have a "catch all", it is better to have it on top of everything else.
That way, when you extract a named field out of it, everything else stays at the same place.

But, rethinking it more in details, I can say also the opposite is true to some extend. 

What it _really_ shows, is that we don't have a clear & explicit convention on those. 😉

penguinpee:
Thank you for taking the time explaining. I see your point now.

Of course, what triggered me submitting this PR was the ever changing colors due to the unordered output. Once I had that sorted, I noticed the "other" value prominently at the top. Pondering your explanation, I'm wondering if the `bind9` plugin needs the "other" field at all. Change the title to "DNS IN queries by type" and leave out "other", being everything that's *not* IN queries.

Then again, I saw that `bind9_rndc` and `bind9` have been merged in master in favor of `bind9_rndc`. That's probably a could decision. On a busy server query logging is preferably turned off since it slows down BIND. I haven't looked at the output of the merged `bind9_rndc`, yet, but plan on doing so when time permits.

Issue #1474, opened 2022-06-06
Cannot run plugin in sandbox environment with systemd-run


penguinpee:
**Describe the bug**
`./dev_scripts/plugin run` fails with:
```
Insecure $ENV{PATH} while running with -T switch at ./devel/munin-monitoring/munin/sandbox/bin/munin-run line 155
```

**To Reproduce**
1. `./dev_scripts/install node`
2. `./dev_scripts/disable_taint`
3. `./dev_scripts/plugin run plugin_name`
4. Above error

**Expected behavior**
Plugin should be executed in sandbox environment.

**Additional context**
I tracked the error down to `dev_scripts/run`. In there `munin-run` is called with `perl -T`. `dev_scripts/disable_taint` only searches `${SANDBOX}/bin`. After changing `dev_scripts/run` I was able to execute my plugin in the sandbox.

steveschnepp:
fixed by your PR #1475

PR #1475, opened 2022-06-06
Possible fix for #1474

penguinpee:
That's what allowed me running plugin in sandbox.

Since taint is disabled inside sandbox, I assume it's okay disabling it in the calling script as well.


sumpfralle:

From my point of view it looks good.

@steveschnepp: what do you think?

PR #1482, opened 2022-07-28
lib/pgsql: fix labels containing #, /, \r or \n

ap-wtioit:
in postgres_size_ it is possible for the datname to contain all
characters supported by postgres. however as per documentation
http://guide.munin-monitoring.org/en/latest/reference/plugin.html
the {fieldname}.label cannot contain # or /. if the label would contain
\r or \n the output of the plugins label would be cut of and or give a
missing label warning and or give lines had errors warning.

Had this issue on one of our servers and currently running the fix there.

Info @wt-io-it

### DIFF
--- a/lib/Munin/Plugin.pm
+++ b/lib/Munin/Plugin.pm
@@ -178,6 +178,28 @@ sub clean_fieldname ($) {
     return $name;
 }
 
+=head3 $label = clean_label($input_label)
+
+Munin plugin field labels are restricted with regards to what
+characters they may use: They cannot contain a
+# (would start a comment),
+\ (no escaping allowed/supported)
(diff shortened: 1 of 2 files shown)


steveschnepp:
Thanks!

Actually, `/` is allowed. It is `\` that isn't. 
As `#` means a comment, and each value has to be on 1 line, `#`, `\n` & `\r` aren't allowed indeed.

That said, would you be able to move it to `Munin::Plugin::clean_label()` instead ?
So, it would not be PostgreSQL specific.

ap-wtioit:
thanks, that makes sense (`\` instead of `/`), will fix it.

and yes i though about adding a common function, will move the code as well.

PR #1487, opened 2022-08-19
cpuspeed: fix average clock frequency calculation

akorn:
The previous code had the formula for the weighted mathematical average
wrong. It added up all `frequency*time` values, then divided by 100. For long
uptimes, this yielded obviously bogus frequency values. The correct formula
is to add up all `frequency*time` values, then divide by the sum of all time
values.

TODO: instead of reporting (only) a running average since the last boot,
save the stats on each run and compute the average for the period since the
last run.


akorn:
Looking at the code more, it sets the field type to DERIVE, not GAUGE, for `acpi-cpufreq`, so maybe it wasn't wrong? I'll have to think about this when I'm less sleepy. :)

akorn:
Yes, I convinced myself the original code is correct. Please disregard this pull request, and sorry about the noise. :)

steveschnepp:
Still, thanks for your patch & its verification!

PR #1502, opened 2022-12-22
cgi/graph: replace Date::Manip with Date::Parse

soliad:
From https://github.com/munin-monitoring/munin/commit/47a2ea75899a6395a80918d043a21d61fe712b44 for the stable branch


yunal:
This replaces one non-core module with another.  Why is Date::Parse better?

steveschnepp:
IIRC, it is that `Date::Manip` does rely on a UNIX external tool, which means forking a process. `Date::Parse` at least doesn't.

This looks like a win to me. Small win, but still a win.

yunal (approved):
Given the info this is a big win.

PR #1511, opened 2023-02-11
exit(0) on --version

lifeofguenter:
`--version` on commands should not exit with an error.

### DIFF
--- a/script/munin-graph
+++ b/script/munin-graph
@@ -85,6 +85,7 @@ sub print_version_and_exit {
     Pod::Usage::pod2usage(
         -verbose  => 99,
         -sections => 'VERSION|COPYRIGHT',
+        -exit     => 0,
     );
 }
 
--- a/script/munin-html
+++ b/script/munin-html
@@ -80,6 +80,7 @@ sub print_version_and_exit {
     Pod::Usage::pod2usage(
         -verbose  => 99,
         -sections => 'VERSION|COPYRIGHT',
+        -exit     => 0,
     );
 }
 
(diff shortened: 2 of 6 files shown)


script/munin-html
```diff
@@ -80,6 +80,7 @@ sub print_version_and_exit {
     Pod::Usage::pod2usage(
         -verbose  => 99,
         -sections => 'VERSION|COPYRIGHT',
+		-exit     => 0,
```
kenyon, line 83:
```suggestion
        -exit     => 0,
```

script/munin-graph
```diff
@@ -85,6 +85,7 @@ sub print_version_and_exit {
     Pod::Usage::pod2usage(
         -verbose  => 99,
         -sections => 'VERSION|COPYRIGHT',
+		-exit     => 0,
```
kenyon, line 88:
```suggestion
        -exit     => 0,
```

.editorconfig
```diff
@@ -0,0 +1,15 @@
+root = true
```
kenyon, line 1: Unrelated change, should be removed from this commit.

.gitignore
```diff
@@ -29,6 +29,7 @@ local/
 # editor files
 .idea/
 .*.sw[op]
+/*.iml
```
kenyon, line 32: Unrelated change, should be removed from this commit.

kenyon (changes requested):

.editorconfig
lifeofguenter: not unrelated, but something I would like to introduce as well :)

.gitignore
lifeofguenter: I see other editor files being ignored as well, why not?

.editorconfig
kenyon: Introduce in a separate pull request then.

.gitignore
kenyon: Should be a separate pull request.

PR #1516, opened 2023-02-26
Switch to rewrite-repeat-if-not-file

kboenke:
Leverage new functionality of mod_rewrite to ensure static content is delivered more reliably. Works on Munin 2.0.57 and lighttpd 1.4.63 under Ubunutu 22.04 x86_64. See also https://redmine.lighttpd.net/projects/lighttpd/wiki/Mod_rewrite#urlrewrite-repeat-if-not-file


steveschnepp:
I edited the commits to remove the commented line. 

I cannot test, does that still work?

kboenke:
Yes, confirmed that this is still working (at least given the versions mentioned before).

PR #1543, opened 2023-06-23
Add update order to nodes

mtillberg:
Add an update_order option to nodes to allow users to specify the order for accessing nodes. 

Nodes should be visited with the longest running nodes first, to minimize the overall munin_update runtime.  I've also included a change to the munin_update plugin to add total runtime so this can be verified.   This is an optional setting, and will have no effect if not used by users.

### DIFF
--- a/doc/reference/munin.conf.rst
+++ b/doc/reference/munin.conf.rst
@@ -375,6 +375,13 @@ only to that node.
    If you make a virtual node which borrow data from real nodes for aggregate graphs, set this to
    "no" for that node.
 
+.. option:: update_order <number>
+
+   Manually define the order that nodes will be handled.  Lower numbers are run first.
+
+   For highest overall throughput, slow running nodes should be run first.  Numbers do not have
+   to be unique, and unspecified nodes are placed at the end of the list.
+
 .. _master-conf-plugin-directives:
 
 PLUGIN DIRECTIVES
--- a/doc/reference/munin.conf.rst
+++ b/doc/reference/munin.conf.rst
@@ -375,9 +375,9 @@ only to that node.
    If you make a virtual node which borrow data from real nodes for aggregate graphs, set this to
    "no" for that node.
 
-.. option:: update_order <number>
+.. option:: update_priority <number>
 
-   Manually define the order that nodes will be handled.  Lower numbers are run first.
+   Manually define the priority that nodes will be handled.  Lower numbers are run first.
 
    For highest overall throughput, slow running nodes should be run first.  Numbers do not have
    to be unique, and unspecified nodes are placed at the end of the list.
--- a/lib/Munin/Common/Config.pm
+++ b/lib/Munin/Common/Config.pm
@@ -138,6 +138,7 @@ my %legal = map { $_ => 1 } qw(
 	unknown
 	unknown_limit
 	update
+	update_order
 	update_rate
 	use_default_name
 	use_node_name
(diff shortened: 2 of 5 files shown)


steveschnepp:
> Nodes should be visited with the longest running nodes 

I'd favor the exact opposite, as if you have the longest running nodes first, you might clog the whole pool and therefore miss many nodes.

So a random shuffling is rather  nice to give every node a fair chance to run. 

Ordering can eventually take the statistics into account, but shuffling also has the nice property of smoothing out any resonance effect. 

That said, I think that giving manual control to the order is a valid feature. 

Yet `order` is a misnomer, I'd prefer the naming `priority` as it does then work exactly how you think it does.

mtillberg:
I've renamed to update_priority.  For the documentation, I'll defer to your expertise.  It may be an issue of small host count vs large or long enough runtime to expect to hit the 5 minute limit.  For posterity, here's the effect on my setup (recent 2.0, 17 hosts, 12 threads) when turning off update ordering for a while:

munin_update-pinpoint-nohosts

For me, the ordering effectively means 2 threads for 2 hosts, and 10 threads for the remaining 15.

steveschnepp:
That graph looks interesting, but I need to understand what is really ongoing 🤔 

Yet, I think I'll merge the PR. I'm also wondering about the best ordering... either priority ASC as you proposed, or priority DESC as I'd expect when not reading the documentation...

steveschnepp:
Also, did you try the `master` branch?
Does it improve?
Does is your PR still making a difference?

mtillberg:
Maybe a simple example would help.  Imagine 40 hosts that take 1 second each, 1 host that takes 10 seconds, and munin server is running with 5 threads.  In the fastest first mode, the 40 hosts would run first at 5 per second for 8 seconds, then the slow host runs for 10 seconds, giving a total of 18 seconds.  In the slowest first mode, the slow host occupies one thread for 10 seconds, but that leaves 4 threads for the remaining 40 hosts, which is a concurrent 10 seconds, so munin_update finishes in 10 seconds.  In random mode, the runtime depends on where the slow host starts.  If it's in position 1-5, it's in the first set of threads and the whole process takes 10 seconds.  If it's in 6-10, it'll be delayed by a second, and the run will take 11 seconds. Starting in position 11-15 will take 12 seconds, etc.

So you shouldn't be worrying about maximizing short term hosts per second.  Maximizing parallelization means to maximize the number of running threads (in this case in a fixed pool size) or alternatively to minimize idling threads.  Absent other constraints, monopolizing a thread with work is a good thing.

If there's a situation where total cumulative node time / thread count is hitting munin_update's 5 minute sampling window, the solution should be to either increase thread count or the time window for the update.  Doing things in any sorted order would be bad, as the same hosts/plugins will generally be missed.  Here, random sampling would make sense, it would be a sort of "sparse" monitoring.

For checking my PR in master, I only checked the the ordering worked.  I'll go ahead and verify the full setup as well.

mtillberg:
Here's the test with master.  15 nodes, only 6 threads in order to make the differences larger.  Started with update ordering disabled, then I turned ordering on.  So the PR is working against master as well. (I lost some early unsorted data, I didn't know the sandbox defaults to 42 data points by default).

munin_update_master

FYI, the timing data is strange. Node times are normal running ./dev_scripts/run munin-update on the command line, but it's much slower under cron.  The shape of the graph suggests a threading issue.

steveschnepp:
I just merged it to `master`. 

Let's improve it from there if needed.

PR #1551, opened 2023-07-02
fix merging field list and graph_order.

mtillberg:
Fix merging plugin field order with config graph_order, and apply after ds changes are in the DB.

### DIFF
--- a/lib/Munin/Master/UpdateWorker.pm
+++ b/lib/Munin/Master/UpdateWorker.pm
@@ -398,19 +398,6 @@ sub _db_service {
 		$self->_db_service_attr($service_id, $attr, $_service_value);
 	}
 
-	# Update the ordering of fields
-	{
-		my @graph_order = split(/ /, $service_attr->{graph_order});
-		DEBUG "_db_service.graph_order: @graph_order";
-		my $ordr = 0;
-		for my $_name (@graph_order) {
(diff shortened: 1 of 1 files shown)


steveschnepp:
Sorry, I don't really understand what you want to do 😞

mtillberg:
This is the munin 2.0 capability where if graph_order was undefined in the config, the order would be inferred from the order the fields are defined in the config.  Basically, the code was there, but the grep was broken.  Then, the updates to ds.ordr are moved to after the ds records are inserted.

As a side note, perl now has randomized hash iteration.  So some plugins may no longer have a stable ordering based on keys(%hash)

yunal:
According to my memory perl never preserved order in hash iteration (I only discovered that this was a thing when I started using Ruby).

steveschnepp:
@mtillberg oh... it does makes sense now. 

@yunal it indeed never *preserved* order, but once inside order was *deterministic* from each run. Starting in 5.18 even the [seed was randomized](https://metacpan.org/pod/perl5180delta#Hash-overhaul). 

Initially for security, but it had the nice sideeffect to expose subtle bugs such as this one.

steveschnepp:
I analyzed it further, and actually you were mostly on spot, it is the merging that is failing.
It was not testing for equality, but for non-nullness 🤦

PR #1552, opened 2023-07-02
prevent malicious config lines

mtillberg:
Filters out config lines that are like:

field.rrd:file /dev/null

I don't think this is a security issue, in testing this causes a unique index violation.  A stricter regex on the line may be a better solution.


steveschnepp:
This sounds sane Indeed. As a matter of fact, we should not accept any `:` in a plugin name at all.

PR #1561, opened 2023-07-25
Sed in the blib directory before install

niclan:
Sed when not making packages resulted in `grep -rl --null "@@" .` which includes _everything_ also the .git directory and one of the git .pack files contained a match for the sed expression thereby corrupting the git checkout.  And also not making any substitutions in the installed files in /usr/local...  So sed has to be done prior to install in the blib directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,11 +51,12 @@ doc:
 
 .PHONY: install
 install: $(BUILD_SCRIPT)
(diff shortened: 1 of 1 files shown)


steveschnepp (approved):

niclan:
@steveschnepp : I'm not allowed to merge, I'm not authorized.

yunal:
I'm also not allowed to merge.

PR #1563, opened 2023-07-25
Make the perl code deterministic

niclan:


steveschnepp:
What do you think about making `graph.pm` and `html.pm` dumb and all the ordering precomputed during update amd stored in the DB?

yunal:
Moving field ordering to the munin-update would work well both for html and graphing. Do you want to give it a go?  Beware: It needs to preserve the order of "config" not "fetch".

The current query in graph.pm does not seem deterministic but I guess I need to get it on a debug test bench :I

Also the graphs seem to be using the default palette, which is really weird to me.

steveschnepp:
> Moving field ordering to the munin-update would work well both for html and graphing. Do you want to give it a go?

That would be great! 

Note that :
* I usually favor ordering directly via the SQL. So priority & ordering is best stored inside the DB as a specific field
* I think there's already some work about it. Not sure about its state, but we should not have too many ways to implement the same thing :wink: 

> Beware: It needs to preserve the order of "config" not "fetch".

That's indeed my understanding also. As `fetch` shall only contain values & `ext_info`, no metadata. There was an issue about what to do with `fetch` only fields, and my advice was to store the data, but not display them. 

That way the config error is much more apparent, yet we don't loose data. The user can still delete the RRD himself if was a config mistake.

> The current query in graph.pm does not seem deterministic but I guess I need to get it on a debug test bench :I

Graph & HTML should be as dumb as possible : even the query should simply leverage `ORDER BY`.
I'm trying to keep any data massaging to a minimal level inside both. That usually makes the DB self explanatory, and ease understanding, as it offers a structured way of the _whole_ state.
 
> Also the graphs seem to be using the default palette, which is really weird to me.

Ah, that's possible, as I might have missed it. Sorry!

niclan:
There is already code to merge graph_order and field_order here https://github.com/munin-monitoring/munin/blob/master/lib/Munin/Master/UpdateWorker.pm#L669

niclan:
This PR is made redundant by #1566

PR #1565, opened 2023-07-26
Resolve ${graph_period} in field labels

niclan:
Also fix misunderstanding wrt rrdname vs label

### DIFF
--- a/lib/Munin/Master/HTML.pm
+++ b/lib/Munin/Master/HTML.pm
@@ -819,6 +819,11 @@ sub _get_params_services {
 sub _get_params_fields {
 	my ($dbh, $service_id) = @_;
 
+	my $sth_sa = $dbh->prepare_cached("SELECT value FROM service_attr WHERE id = ? and name = ?");
+
+	$sth_sa->execute($service_id, 'graph_period');
+	my ($graph_period) = $sth_sa->fetchrow_array();
+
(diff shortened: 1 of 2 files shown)


lib/Munin/Master/HTML.pm
```diff
@@ -819,6 +819,11 @@ sub _get_params_services {
 sub _get_params_fields {
 	my ($dbh, $service_id) = @_;
 
+	my $sth_sa = $dbh->prepare_cached("SELECT value FROM service_attr WHERE id = ? and name = ?");
+
+	$sth_sa->execute($service_id, 'graph_period');
```
niclan, line 824: @steveschnepp : any clue to why this prepare/execute returns a empty value?

Compare https://github.com/munin-monitoring/munin/blob/master/lib/Munin/Master/Graph.pm#L227 and the ds query further down.

lib/Munin/Master/HTML.pm
```diff
@@ -819,6 +819,11 @@ sub _get_params_services {
...
+	my ($graph_period) = $sth_sa->fetchrow_array();
```
niclan, line 825: @steveschnepp : any clue to why this prepare/execute returns a empty value?

Compare https://github.com/munin-monitoring/munin/blob/master/lib/Munin/Master/Graph.pm#L227 and the ds query further down.

PR #1567, opened 2023-07-26
Get the colours right

niclan:


steveschnepp:
Ha. I do not remember why I used the `gfx` namespace. 🤦 

Good catch!

yunal:
You merge?

PR #1577, opened 2023-09-14
upgrade jQuery to 3.7.1

kenyon:
This resolves security vulnerabilities in jQuery.

The previous upgrade was done in #1380.


steveschnepp:
Awesome!

Minor nitpick, for munin I usually prefer the non-minified version. As we do not really have huge traffic, do we? ;)

contrib/plugin-gallery/static/gallery-footer.html
```diff
@@ -4,7 +4,7 @@
 	<footer>
 		<p class="tagline">This page was generated by <a href="http://munin-monitoring.org/wiki/PluginGallery">gallery-build.sh</a>.</p>
 	</footer>
-	<script src="/static/js/jquery-1.12.4.js"></script>
+	<script src="/static/js/jquery-3.7.1.min.js"></script>
```
steveschnepp, line 7: Can we have the non-minified version ?

steveschnepp (approved):

contrib/plugin-gallery/static/gallery-footer.html
kenyon: On https://jquery.com/download/ they call them "production" versus "development", so I went with "production". 🤷‍♀️

PR #1592, opened 2023-11-08
Shared memory does not included in Cached memory on Linux.

starinacool:


starinacool:
https://www.kernel.org/doc/html/latest/filesystems/proc.html#meminfo

kenyon:
Are you sure? See #554.

PR #1601, opened 2023-11-29
Update `ipmi_sensor_` to handle ipmitool `Unable to read sensor` errors

smokris:
Occasionally (once every few days) `ipmitool` will output a reading like this:
```
Sensor Reading        :  Unable to read sensor: Device Not Present
```

Currently, Munin's `ipmi_sensor_` produces the following output in that situation:
```
fan1.value Unable
```
…which causes Munin to log:
```
Failed to interpret expected numeric value of field 'fan1' (host 'palo.kosada.com'): 'Unable'
Argument "Unable" isn't numeric in sprintf
```
…and issue alerts as though the reading had a value of `0`.

This PR modifies `ipmi_sensor_` to omit the invalid value, which should avoid the unnecessary alerts.

### DIFF
--- a/plugins/node.d/ipmi_sensor_
+++ b/plugins/node.d/ipmi_sensor_
@@ -100,6 +100,9 @@ def parse_data(data):
      Assertions Enabled    : lcr-
      Deassertions Enabled  : lcr-
 
+    When a sensor reading is unavailable, ipmitool outputs:
+     Sensor Reading        :  Unable to read sensor: Device Not Present
+
     """
     sensors = {}
     cur_sensor = None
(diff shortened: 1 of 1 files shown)


kenyon (changes requested):
Returning a value of `U` is the correct thing to do in this situation, which means an unknown/undefined value at that time. This is documented in https://guide.munin-monitoring.org/en/latest/develop/plugins/plugin-concise.html#fetch.

smokris:
Ah, thanks.  I just amended it to output `fan1.value U` when the sensor reading is unavailable.

kenyon (approved):

PR #1618, opened 2024-05-13
Fix cpuspeed for low loads

### LINKED ISSUE #1617: cpuspeed plugin problems
Dear munin developers and mainainers! I would like to share my observations and solutions with you about cpuspeed plugin.

**Describe the bug**

1. Graph shows nan value when frequency is under minimum. Example: minimum is 700Mhz, my value: 699964hz.
2. nan value is misleading! It is a number, just bellow minimum value.
3. The CPU intensity of munin update procedure raises the frequency over the normal value. So it fakes the measurement and graph is useless by default (see attached images).

**Expected behavior**

1. Allow values bellow minimum like in case of: https://github.com/munin-monitoring/munin/commit/76fe8bf5fc0a1830b967a5b1dc7e68e0060ac4f9
So adding `MINHZ=$(( $MINHZ - $MINHZ / 10 ))` [here](https://github.com/munin-monitoring/munin/blob/47517b38cb11e924ca238bfe2626113071718eb1/plugins/node.d.linux/cpuspeed#L157).
2. Simple dash character or "oor" (out of range) would be more helpful to debug.
3. I have added in case of Intel CPU `sleep 0.2` before measuring CPU frequencies in the script ([here](https://github.com/munin-monitoring/munin/blob/47517b38cb11e924ca238bfe2626113071718eb1/plugins/node.d.linux/cpuspeed#L167)). In my case this 200ms is enough for the Intel N100 CPU to slow down to normal frequency. For me that is an acceptable delay.

**To Reproduce**
Just have an idle system with Alder Lake or newer Intel CPU.

**Screenshots & Logs**
nan value when frequency is bellow minimum. I had to zoom in to provide proof:
nan
Graph by default:
default_mess
After adding 200ms sleep before measurement and fixing minhz:
after_sleep

**Environment**
 - OS+Distribution Version: Ubuntu 23.10
 - Kernel: 6.5.0-28-generic x86_64
 - Munin Version: 2.0.71-2ubuntu1

**Additional context**
I have an Intel N100 CPU which changes frequency quite often (plenty time within a second). My setup is idle in most of the time, so mostly the CPU is running in minimum frequency.

steveschnepp:
This contains 2 fixes for low loaded nodes that have very dynamic frequency.

Closes #1617

### DIFF
--- a/plugins/node.d.linux/cpuspeed
+++ b/plugins/node.d.linux/cpuspeed
@@ -18,7 +18,19 @@ None needed by default.
 You can set one environment variable to modify the plugins behaviour:
 
   [cpuspeed]
+     env.sleep .2
      env.scaleto100 yes
+
+Parameter sleep 
+
+Sleep for some (fractional) seconds before actually taking the measure.
+This is useful to enable the CPU to relax a little from the previous plugins.
+
+"env.sleep .2" will cause a 200ms sleep.
+
+Default is to NOT sleep, aka "env.sleep 0".
+
+Parameter scaleto100
 
 Show the frequency as a percentage instead of absolute frequency.
 If set the "yes" the plugin sets up a CDEF to change the speed in Hz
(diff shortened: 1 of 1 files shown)


baranyaib90 (approved):
Perfect, thank you very much!

Issue #1641, opened 2024-12-31
[Bug] Munin-node fails to start when 'host *' is set in munin-node.conf


leemgs:
**Describe the bug**
The host * in /etc/munin/munin-node.conf can cause errors in munin-node. 

I downloaded munin/munin-node from ubuntu 24.04 as shown below and installed it after compiling. 
Unfortunately, the “host *” statement in the munin-node.conf file is causing munin-node to fail to run properly. 

https://github.com/munin-monitoring/munin/blob/master/etc/munin-node.conf.PL#L81

Munin-node uses a Perl module called Net::Server internally to handle network connections. In the Net::Server module, the host setting must explicitly specify a specific IP address or hostname. When Net::Server actually interprets the * wildcard, it repeatedly attempts to bind all available network interfaces on the system. This process can result in crashes or errors.

**To Reproduce**
1. sudo apt install munin-node
2. sudo vi /etc/default/munin-node
    DAEMON_ARGS="--config /etc/munin/munin-node.conf"
3.  vi /etc/munin/munin-node.conf
    host *
4. sudo systemctl start  munin-node
5. sudo systemctl status munin-node
6. See below errors
```

root@u2404:/etc/munin# sudo systemctl start  munin-node
Job for munin-node.service failed because the control process exited with error code.
See "systemctl status munin-node.service" and "journalctl -xeu munin-node.service" for details.
root@u2404:/etc/munin#
root@u2404:/etc/munin#
root@u2404:/etc/munin# sudo systemctl status munin-node
× munin-node.service - Munin Node
     Loaded: loaded (/usr/lib/systemd/system/munin-node.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Tue 2024-12-31 18:22:04 KST; 1s ago
   Duration: 4min 50.784s
       Docs: man:munin-node(1)
             http://guide.munin-monitoring.org/en/stable-2.0/reference/munin-node.html
    Process: 280164 ExecStartPre=/usr/bin/install -o munin -g munin -d /run/munin (code=exited, status=0/SUCCESS)
    Process: 280166 ExecStart=/usr/sbin/munin-node --foreground $DAEMON_ARGS (code=exited, status=1/FAILURE)
   Main PID: 280166 (code=exited, status=1/FAILURE)
        CPU: 140ms

12월 31 18:22:04 u2404 systemd[1]: munin-node.service: Scheduled restart job, restart counter is at 5.
[...]

PR #1665, opened 2025-09-27
fix Issue #1641 'Munin-node fails to start when 'host *' in config

jpoehlmann:
In the default configuration file /etc/munin/munin-node.conf

"host *" 

does no longer work (seen in debian trixie). Issue #1641 states that.

Replacing "*" by 0.0.0.0 lets the munin node start and accept connections on all interfaces including localhost and so effectively fixes Issue #1641.

I changed this setting in the default node-config and a config documentation file.

Background: see eg.g https://www.baeldung.com/cs/ip-all-zeroes

### DIFF
--- a/doc/reference/munin-node.conf.rst
+++ b/doc/reference/munin-node.conf.rst
@@ -208,7 +208,7 @@ A pretty normal configuration file:
   # cidr_deny  192.0.2.42/32
 
   # Which address to bind to;
-  host *
+  host 0.0.0.0
   # host 127.0.0.1
 
   # And which port
--- a/etc/munin-node.conf.PL
+++ b/etc/munin-node.conf.PL
@@ -78,7 +78,7 @@ allow ^::1$
 
 # Which address to bind to.
 #host 127.0.0.1
-host *
+host 0.0.0.0
 
 # The TCP port the munin-node listens on.
 port 4949


doc/reference/munin-node.conf.rst
```diff
@@ -208,7 +208,7 @@ A pretty normal configuration file:
   # cidr_deny  192.0.2.42/32
 
   # Which address to bind to;
-  host *
+  host 0.0.0.0
```
kenyon, line 211:
```suggestion
  host ::
```

etc/munin-node.conf.PL
```diff
@@ -78,7 +78,7 @@ allow ^::1$
 
 # Which address to bind to.
 #host 127.0.0.1
-host *
+host 0.0.0.0
```
kenyon, line 81:
```suggestion
host ::
```

kenyon (changes requested):
I don't think 0.0.0.0 is going to work with IPv6. Can you try ::?

etc/munin-node.conf.PL
jpoehlmann: Sorry, formatting error.

I propose the following configuration template:

      # activate only one "host" statement, comment out the other  
      #bind to IP V6
      host *
      #bind to IP V4
      #host 0.0.0.0

kenyon:
Replaced by #1666.

PR #1674, opened 2025-11-29
plugins/nut_*: emit 'U' if value is not seen in upsc output

tofurky:
Not all UPS support all metrics that the plugins look for.

In my case, some CyberPower brand units don't show input_frequency/ups_temperature for nut_misc, and
input_voltage_minimum/input_voltage_maximum for nut_volts.

To prevent generating log messages about invalid lines during each update, emit a value of 'U' for the missing attribute(s).

### DIFF
--- a/plugins/node.d/nut_misc.in
+++ b/plugins/node.d/nut_misc.in
@@ -100,7 +100,7 @@ sub fetch_values {
 	}
 	$status{'battery_runtime'} /= 60;
 	foreach my $label (sort keys %graph) {
-		print "$label.value $status{$label}\n";
+		print "$label.value " . ($status{$label} // 'U') . "\n";
 	}
 }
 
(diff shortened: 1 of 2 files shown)


tofurky:
I have a commit ready for master as well - however, AFAICT it *won't* actually complain (at least in the same place, with the same message) about the invalid lines there. Could rework the commit msg?

https://github.com/munin-monitoring/munin/compare/master...tofurky:munin:nut_missing_values

tofurky:
Also, didn't think about it at the time, but the `//` undef coalescing operator requires Perl 5.10 circa 2007; I see some _old_ notes about removing the usage of it in ChangeLog `munin (1.4.0-alpha, r1560:2898, 2009-11-06). Probably fine at this point? cpanfile references `v5.10.1` and has been there since 2018.

